PR #89, opened 2016-06-13
[WIP]CON-3351 Added sku in export dql query

mkafadarov:


mkafadarov:
This should be merge after CON-3121 is done

PR #90, opened 2016-06-13
[CON-3272] fix redirect url

svenfinke:


mkafadarov:
@svenfinke why did you use **empty($configComponent->getConfig('connectDebugHost', ''))** ? 
Is **Config::getConfig** method returns the default value if the result is empty ?

svenfinke:
@mkafadarov The default value was never used, even if just an empty string was saved in the config. Therefore I used "empty()" and to make sure that it always works - even if the default behaviour is "fixed" in future shopware versions - I have set the default of getConfig to ''.

PR #92, opened 2016-06-16
CON-3171 price workflow

sbofirov:


Controllers/Backend/Connect.php
```diff
@@ -542,7 +542,7 @@ public function loginAction(){
         if ($host) {
             $host = 'sn.' . $host;
         } else {
-            $host = $this->getConfigComponent()->getConfig('marketplaceNetworkUrl');
+            $host = 'sn.' . $this->getConfigComponent()->getMarketplaceUrl();
```
mkafadarov: Please use const instead of magic string

PR #96, opened 2016-06-21
CON-3272 Added auto login to social network system

mkafadarov:


Controllers/Backend/ConnectBaseController.php
```diff
@@ -24,6 +24,7 @@
 
 namespace ShopwarePlugins\Connect\Controllers\Backend;
 
+use Firebase\JWT\JWT;
```
sbofirov: please remove

PR #101, opened 2016-06-27
CON-3380 Fixed creation of category on dropping remote category to lo…

mkafadarov:
…cal tree

### DIFF
--- a/Components/CategoryResolver/AutoCategoryResolver.php
+++ b/Components/CategoryResolver/AutoCategoryResolver.php
@@ -25,11 +25,6 @@ class AutoCategoryResolver implements CategoryResolver
      */
     private $remoteCategoryRepository;
 
-    /**
-     * @var array()
-     */
-    private $leafCollection = array();
-
     public function __construct(
         ModelManager $manager,
         CategoryRepository $categoryRepository,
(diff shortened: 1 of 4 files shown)


Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -155,6 +166,22 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+    {
+        $this->leafCollection[] = $leaf;
+    }
+
+    /**
+     * @return Category[]
+     */
+    public function getLeafCollection()
+    {
+        return $this->leafCollection;
```
sbofirov: Please refactor this method, ```public function getLeafCollection(array $tree)```
store leaf categories in array and return them.
You don't need to use property for that.

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -110,6 +119,8 @@ public function convertTreeToEntities(array $tree, $parent = null)
 
             if (!empty($category['children'])) {
                 $this->convertTreeToEntities($category['children'], $categoryModel);
+            } else {
+                $this->addLeaf($categoryModel);
```
sbofirov: This can be removed after refactoring.

Models/Connect/ProductToRemoteCategoryRepository.php
```diff
@@ -109,4 +109,29 @@ public function findArticleIdsByRemoteCategory($remoteCategoryKey)
...
         }, $result);
     }
+
+    /**
+     * Collects article ids by given
+     * remote category name
+     * @param string $remoteCategoryName
+     * @return array
+     */
+    public function findArticleIdsByRemoteCategoryName($remoteCategoryName)
```
sbofirov: This is not needed anymore. You can use ```findArticleIdsByRemoteCategory```, just needs modifications.

PR #116, opened 2016-07-20
[CON-3539] fix the menu creation after logging in or registering

svenfinke:


Bootstrap/Setup.php
```diff
@@ -103,7 +103,7 @@ private function createMyMenu()
                     'label' => 'Export',
                     'controller' => 'Connect',
                     'action' => 'Export',
-                    'class' => 'sc-icon-export',
+                    'class' => 'contents--import-export',
```
mkafadarov: This class is the same as import, should be different for the export ?

mkafadarov: I think you should change the class in 
`$db->exec(strtr($insertSql, [
             '#parent#' => $row['parent'],		             '#parent#' => $row['parent'],
             '#name#' => 'Import',		             '#name#' => 'Import',
             '#class#' => 'contents--import-export',		             '#class#' => 'contents--import-export',
             '#pluginID#' => $row['pluginID'],		             '#pluginID#' => $row['pluginID'],
             '#controller#' => 'Connect',		             '#controller#' => 'Connect',
             '#onclick#' => '',		             '#onclick#' => '',
             '#action#' => 'Import'		             '#action#' => 'Import'
         ]));		         ]));` with sc-icon-export and sc-icon-import

PR #126, opened 2016-08-01
[CON-3502, CON-3503] change wording and rearrange fields

svenfinke:


Snippets/backend/connect/view/main.ini
```diff
@@ -480,9 +483,10 @@ config/log_label = "Logging aktivieren"
 config/success/message = "Änderungen erfolgreich übernommen"
 config/import/pictures_label = "Bilder beim Produkt-Erstimport importieren"
 config/import/pictures_help = "Der Import von Bildern kann den Import verlangsamen. Wenn Sie viele Produkte importieren möchten, sollten Sie diese Option nicht aktivieren und die Bilder über den CronJob oder »Geänderte Produkte« importieren."
-config/import/overwrite_properties = "Folgende Felder beim Produkt-Aktualisierung überschreiben"
+config/import/overwrite_properties = "Felder, die bei Artikel-Aktualisierungen überschrieben werden"
+config/import/overwrite_properties_helptext = "Gebe an, welche Felder überschrieben werden sollen, wenn dein Lieferant sie ändert. Diese Einstellung kannst du auch pro Artikel treffen. Gehe dafür direkt in den Artikel und dann auf den Tab Connect."
 config/import/overwrite_properties_help = "Die hier ausgewählten Felder werden automatisch überschrieben, wenn der Quellshop diese ändert. Sie können auf Artikel-Ebene Ausnahmen definieren."
-config/import/overwrite_product_name = "Produkt-Name"
+config/import/overwrite_product_name = "Artikelname"
```
sbofirov: You need to remove all changed snippets or just to use different keys, because I still see "Produkt-Name" instead of "Artikelname".

Same for:
* config/import/overwrite_properties = "Folgende Felder beim Produkt-Aktualisierung überschreiben"
* config/import/product_import_settings_title = "Produktbeschreibungen"
* config/import/image_settings_title = "Produktbilder"

svenfinke: But on reinstall the snippets are being replaced... Should we really make it complicated even though the plugin has not been released? I know that we need to do it in a different way after release, but at the moment it is fine like it is, isn't it?

sbofirov: @svenfinke I didn't reinstall the plugin, then just add a comment in your ticket, how this change should be tested. Just to know that reinstall is required.

PR #127, opened 2016-08-01
CON-3494 Generated change for make main variant

mkafadarov:


Components/ProductToShop.php
```diff
@@ -755,4 +755,18 @@ public function changeAvailability($shopId, $sourceId, $availability)
...
     }
+
+    public function makeMainVariant($shopId, $sourceId, $groupId)
+    {
+        // find article detail id
+        $result = $this->manager->getConnection()->fetchAssoc(
+            'SELECT article_id, article_detail_id FROM s_plugin_connect_items WHERE source_id = ? AND shop_id = ?',
+            array($sourceId, $shopId)
+        );
+
```
sbofirov: Please check if ```$result``` is empty

PR #132, opened 2016-08-05
CON-3321 Categories in import are kept expanded

StefanStefanoff:


mkafadarov:
- [x] Reviewed (@mkafadarov)

PR #134, opened 2016-08-10
CON-3554 Added connect categories deactivation

mkafadarov:


Components/CategoryExtractor.php
```diff
@@ -95,6 +98,38 @@ public function extractImportedCategories()
...
 
     /**
+     * @param $parentId
+     * @return array
+     */
+    public function getCategoryIdsCollection($parentId)
+    {
+        $categoryRepository = $this->categoryResolver->getCategoryRepository();
+
+        $mainCategory = $categoryRepository->findOneBy(array('id' => $parentId));
```
sbofirov: ```$mainCategory``` always exists ?

Components/CategoryExtractor.php
```diff
@@ -95,6 +98,38 @@ public function extractImportedCategories()
     }
 
     /**
+     * @param $parentId
+     * @return array
+     */
+    public function getCategoryIdsCollection($parentId)
+    {
+        $categoryRepository = $this->categoryResolver->getCategoryRepository();
```
sbofirov: ```getCategoryRepository``` method is not defined in CategoryResolver interface. I think it's not good idea to define method like this in interface. I'd suggest to inject CategoryRepository in CategoryExtractor.

Btw please check ConnectFactory on line 137, when we use ```AutoCategoryResolver``` and when ```DefaultCategoryResolver```.  It depends on createCategoriesAutomatically configuration.

Controllers/Backend/Import.php
```diff
@@ -300,15 +300,41 @@ public function activateArticlesAction()
...
+        if (!$categoryId) {
+            return $this->View()->assign(array(
+                'success' => false,
+                'error' => 'Please select a category for deactivation',
+            ));
+        }
+
+        $categoryIds = $this->getCategoryExtractor()->getCategoryIdsCollection($categoryId);
+
+        if (count($categoryIds) > 1) {
```
sbofirov: why > 1 ? can't be 0 in some cases ?

Views/backend/connect/_resources/styles/connect-styles.css
```diff
@@ -85,7 +85,7 @@ table[id^=base-element-fieldset] tbody + tbody .x-btn {
 .tree-table-toolbar {
     border-top: 1px solid #a3b5c1 !important;
     border-right: 1px solid #a3b5c1 !important;
-    border-left: 1px solid #a3b5c1 !important;
+    /*border-left: 1px solid #a3b5c1 !important;*/
```
sbofirov: if not necessary please remove it

Views/backend/connect/view/import/own_categories.js
```diff
@@ -150,6 +150,13 @@ Ext.define('Shopware.apps.Connect.view.import.OwnCategories', {
             dock: 'top',
             ui: 'shopware-ui',
             items: [
+                {
+                    xtype: 'button',
+                    iconCls: 'sprite-minus-circle-frame',
+                    action: 'deactivateCategory',
+                    margin: '0 10px 0 0',
+                    text: '{s name=import/action/deactivate_category}Deactivate categories{/s}'
```
sbofirov: Why "Deactivate categories" ? I think it should be "Deactivate category", because only one category can be deactivated ->
```
 params: {
  categoryId: selected[0].get('id')
 },
```

mkafadarov: If there is a sub category it will be also deactivated

PR #145, opened 2016-08-19
Con 3712 exclude inactive products

sbofirov:


Components/ConnectExport.php
```diff
@@ -97,12 +97,22 @@ public function export(array $ids, ProductStreamsAssignments $streamsAssignments
...
+                $this->manager->flush($connectAttribute);
+                continue;
+            }
 
-            $prefix = $item['title'] ? $item['title'] . ' ('. $item['number'] .'): ' : '';
-            if (empty($item['exportStatus'])
-                || $item['exportStatus'] == Attribute::STATUS_DELETE
-                || $item['exportStatus'] == Attribute::STATUS_ERROR
-            ) {
+            if ($this->helper->isProductExported($connectAttribute)) {
```
StefanStefanoff: @sbofirov Shouldn't this be ``if (**!**$this->helper->isProductExported($connectAttribute))`` since if it is already exported the status should be "Update" ? or am I wrong?

Subscribers/Article.php
```diff
@@ -148,7 +148,7 @@ public function generateMainVariantChange($detailId)
             return;
         }
         // Check if entity is a connect product
-        if (!$attribute->getExportStatus() || $attribute->getExportStatus() == Attribute::STATUS_DELETE) {
+        if (!$this->getHelper()->isProductExported($attribute)) {
```
StefanStefanoff: @sbofirov If we have ``Attribute::STATUS_ERROR_PRICE`` isProductExported() will return false. Is it possible for the product to have been exported and then on Update the price fails and the status is set to STATUS_ERROR_PRICE and is this considered a connect product in this case ?

PR #148, opened 2016-08-24
[CON-3718] only open tab if login or registration were successful

svenfinke:


Views/backend/connect/controller/main.js
```diff
@@ -478,6 +473,7 @@ Ext.define('Shopware.apps.Connect.controller.Main', {
 
                 response.shopwareId = params.shopwareID;
                 me.splashScreen.close();
+                me.openLink('http://google.com');
```
StefanStefanoff: Do we need to redirect to Google here ?

PR #151, opened 2016-08-26
make the plugin work if it is started without reloading.

svenfinke:


Views/backend/connect/controller/main.js
```diff
@@ -120,8 +120,33 @@ Ext.define('Shopware.apps.Connect.controller.Main', {
...
+                    me.launchAction();
+                    me.setEventListeners();
+                }
+            );
+        } else {
+            me.launchAction();
+            me.setEventListeners();
+        }
 
+        me.callParent(arguments);
```
StefanStefanoff: Since Ajax request may be executed after ``me.callParent(arguments)`` is it ok if that happens

svenfinke: As the Plugin Team said, this is okay...

PR #161, opened 2016-09-12
CON-3494 Added flag isEvent in export

mkafadarov:


sbofirov:
:+1:  this is temporary solution, we'll implement it in better way in the future.

PR #170, opened 2016-09-23
[CON-3706] use own connect attribute field

svenfinke:
Yeah... best branch naming typo ever :( :D

-> https://jira.shopware.com/browse/CON-3706


Bootstrap/Update.php
```diff
@@ -223,4 +224,18 @@ private function clearConfigCache()
...
     }
+
+    private function updateConnectAttribute(){
+        if (version_compare($this->version, '1.0.6', '<=')) {
+            $result = Shopware()->Db()->query("SELECT value FROM s_plugin_connect_config WHERE name = 'connectAttribute'");
+            $row = $result->fetch();
+            if ($row) {
+                Shopware()->Db()->query("
+                    UPDATE `s_articles_attributes` 
+                    SET `connect_reference` = `attr" . $row['value'] . "`
```
sbofirov: by default $row['value'] will be = 19. Can be between 1 and 20.
Why you need to set connect_reference value "attr19" for example?
I think we need to fetch which attribute is used right now, 19 for example, then select value from attr19 and insert it in connect_reference column.

connect_reference contains sourceId only for imported products. So this query must be changed.

Bootstrap/Update.php
```diff
@@ -223,4 +224,18 @@ private function clearConfigCache()
             $cacheManager->clearConfigCache();
         }
     }
+
+    private function updateConnectAttribute(){
+        if (version_compare($this->version, '1.0.6', '<=')) {
```
sbofirov: We have to discuss about versions and different branches. It's possible to have another branch with update.

Components/Helper.php
```diff
@@ -108,7 +108,7 @@ public function getArticleModelByProduct(Product $product, $mode = Query::HYDRAT
         return null;
     }
 
-    public function getArticleDetailModelByProduct(Product $product, $mode = Query::HYDRATE_OBJECT)
+    public function  getArticleDetailModelByProduct(Product $product, $mode = Query::HYDRATE_OBJECT)
```
sbofirov: please remove not needed space

Views/backend/connect/view/config/general/form.js
```diff
@@ -396,63 +395,6 @@ Ext.define('Shopware.apps.Connect.view.config.general.Form', {
...
-        });
-
-        var defaultContainer = Ext.create('Ext.container.Container', {
-            columnWidth: 1,
-            layout: 'anchor',
-            border: false,
-            items: [
-                {
-                    xtype: 'combo',
-                    name: 'connectAttribute',
```
sbofirov: please remove "connectAttribute" from Shopware.apps.Connect.model.config.General as well.

Bootstrap/Update.php
```diff
@@ -223,4 +224,18 @@ private function clearConfigCache()
...
+            $result = Shopware()->Db()->query("SELECT value FROM s_plugin_connect_config WHERE name = 'connectAttribute'");
+            $row = $result->fetch();
+            if ($row) {
+                Shopware()->Db()->query("
+                    UPDATE `s_articles_attributes` 
+                    SET `connect_reference` = `attr" . $row['value'] . "` 
+                    WHERE connect_reference IS NULL;
```
sbofirov: After that we're able to remove "connectAttribute" from s_plugin_connect_config table.

Bootstrap/Update.php
svenfinke: But this way I set connect_reference to the value from the attribute field that is used. By default this will be attr19. Not "attr19" is written into the reference field, but the value of attr19.

sbofirov: yeah sorry, my fault

Bootstrap/Update.php
svenfinke: Done.

Components/Helper.php
svenfinke: Done.

Views/backend/connect/view/config/general/form.js
svenfinke: Done.

Bootstrap/Update.php
svenfinke: But this is only relevant for us, isn't it? So this should be fine

sbofirov: I also changed plugin version to 1.0.6, but my branch is not merged yet. So your has higher priority :)

PR #177, opened 2016-10-03
CON-3868 Created listener for stream deletion

mkafadarov:


Components/ProductStream/ProductStreamService.php
```diff
@@ -191,6 +191,11 @@ public function getList($start = null, $limit = null)
         );
     }
 
+    public function isStreamExported($streamId)
+    {
+        return (bool) $this->streamAttrRepository->findOneBy(array('streamId' => (int) $streamId));
```
sbofirov: Please check if the status is export. If it's error or delete, stream is not exported.

Subscribers/ProductStreams.php
```diff
@@ -0,0 +1,43 @@
...
+    public function getSubscribedEvents()
+    {
+        return array(
+            'Enlight_Controller_Action_PostDispatch_Backend_ProductStream' => 'extendBackendProductStream',
+        );
+    }
+
+    public function getProductStreamService()
+    {
+        return $this->Application()->Container()->get('swagconnect.product_stream_service');
```
sbofirov: Please inject stream service as argument in constructor.

PR #178, opened 2016-10-06
CON-3916 Handle error when verification fails on login/registraion

mkafadarov:


Controllers/Backend/ConnectBaseController.php
```diff
@@ -626,27 +624,34 @@ public function loginAction(){
...
             $this->getConfigComponent()->setConfig('apiKey', $responseObject->apiKey, null, 'general');
-            $this->removeConnectMenuEntry();
             $this->getSDK()->verifySdk();
             $this->getConfigComponent()->setConfig('apiKeyVerified', true);
             $this->getConfigComponent()->setConfig('shopwareId', $shopwareId);
+            $this->removeConnectMenuEntry();
             $marketplaceSettings = $this->getSDK()->getMarketplaceSettings();
             $this->getMarketplaceApplier()->apply(new MarketplaceSettings($marketplaceSettings));
+        } catch (\Exception $e) {
+            $this->getConfigComponent()->setConfig('apiKey', null, null, 'general');
```
sbofirov: please move set api key after ```$this->getSDK()->verifySdk();``` then you don't need to set null if something went wrong.

Controllers/Backend/ConnectBaseController.php
```diff
@@ -682,26 +687,34 @@ public function registerAction(){
...
             $this->getConfigComponent()->setConfig('apiKey', $responseObject->apiKey, null, 'general');
             $this->getSDK()->verifySdk();
             $this->getConfigComponent()->setConfig('apiKeyVerified', true);
             $this->getConfigComponent()->setConfig('shopwareId', $shopwareId);
+            $this->removeConnectMenuEntry();
             $marketplaceSettings = $this->getSDK()->getMarketplaceSettings();
             $this->getMarketplaceApplier()->apply(new MarketplaceSettings($marketplaceSettings));
-            $this->removeConnectMenuEntry();
+        } catch (\Exception $e) {
+            $this->getConfigComponent()->setConfig('apiKey', null, null, 'general');
```
sbofirov: same here

PR #185, opened 2016-10-13
[CON-3928] throw new events in SwagConnect Plugin

svenfinke:


Components/ConnectExport.php
```diff
@@ -37,6 +38,9 @@ class ConnectExport
     /** @var ErrorHandler */
     protected $errorHandler;
 
+    /** @var  ContainerAwareEventManager */
+    protected $eventHandler;
```
sbofirov: where it's used ?

svenfinke: It's not used anywhere.. I removed this.

PR #186, opened 2016-10-13
CON-3898 - Renamed connect host option label

Exoridus:
- Removed old option label
- Fixed snippets
- Formatted js code


Exoridus:
https://jira.shopware.com/browse/CON-3898

PR #187, opened 2016-10-14
Remove deprecated function calls

SimonBaeumer:
Removed method calls on deprecated methods.
- Set minimum version to `5.2.0`
- Usage of the new attribute system in `Setup.php` & `Uninstall.php`
- Removed `Shopware()->Bootstrap` calls
- Removed `Enlight()` calls with `Shopware()`

**Todo:**
- `engine/Shopware/Plugins/Local/Backend/SwagConnect/Bootstrap/Setup.php:610`

### DIFF
--- a/Tests/Shopware/Connect/Component/ImportServiceTest.php
+++ b/Tests/Shopware/Connect/Component/ImportServiceTest.php
@@ -54,7 +54,7 @@ class ImportServiceTest extends ConnectTestHelper
         $this->importService = new ImportService(
             $this->manager,
             Shopware()->Container()->get('multi_edit.product'),
-            $this->remoteCategoryRepository,
+            $this->categoryRepository,
             $this->articleRepository,
             $this->remoteCategoryRepository,
             $this->manager->getRepository('Shopware\CustomModels\Connect\ProductToRemoteCategory'),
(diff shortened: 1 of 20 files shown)


Tests/Shopware/Connect/Component/ImportServiceTest.php
```diff
@@ -54,7 +54,7 @@ public function setUp()
         $this->importService = new ImportService(
             $this->manager,
             Shopware()->Container()->get('multi_edit.product'),
-            $this->remoteCategoryRepository,
+            $this->categoryRepository,
```
SimonBaeumer: @svenfinke This crashed on php7

SimonBaeumer:
@svenfinke The `billing` and `shipping` models will be available until Shopware 5.4, so it is not needed to replace the usage of them. 

The only  deprecation which must be fixed is the `\Shopware\Models\Article\Element`

mkafadarov:
@SimonBaeumer, @sbofirov fixed all tests for php7 so you can see his change over here:
https://github.com/shopware/SwagConnect/pull/191

sbofirov:
The changes from this PR are already applied or they will be with following PR https://github.com/shopware/SwagConnect/pull/238

PR #196, opened 2016-10-31
CON-3970 Fixed problem with backend session on uninstalling

mkafadarov:


mkafadarov:
Somehow the Shop repo is breaking the serialized object in 's_core_sessions_backend' table

PR #214, opened 2016-12-14
CON-4042 Price type reset

mkafadarov:


Controllers/Backend/ConnectConfig.php
```diff
@@ -871,6 +928,21 @@ public function getFactory()
     }
 
     /**
+     * @return string
+     */
+    protected function getHost()
```
sbofirov: Please move this function in ConfigComponent. After that you can use it here and in ConnectBaseController as well.
It should be also removed from ConnectBaseController.

Modify in all places where it's called.

Snippets/backend/connect/view/main.ini
```diff
@@ -182,6 +182,10 @@ mapping/columns/connect-category = "[0] Category"
 mapping/columns/shopware-category = "Shopware Category"
 navigation/products = "Products"
 import/toolbar/show_in_groups = "Show in groups"
+config/price_reset_btn = "reset"
+config/price_reset_label = "Reset exported prices"
+config/price_reset_message = "Your exported products will be deleted in Connect and your sent offers will be invalid.<br>Do you want to continue?"
+config/price_reset_success = "The exported prices were successfully reset. It will take up to 10min for the changes to take effect. When this operation is done, you will get the option to set the price type again when you reopen "Export" in the Connect menu."
```
sbofirov: Maybe double quotes ```"Export"``` will break the snippets

Snippets/backend/connect/view/main.ini
```diff
@@ -514,6 +518,10 @@ navigation/mapping/export = "Export"
 mapping/columns/connect-category = "[0] Kategorie"
 mapping/columns/shopware-category = "Shopware Kategorie"
 navigation/products = "Produkte"
+config/price_reset_btn = "zurücksetzen"
+config/price_reset_label = "Exportierte Preise zurücksetzen"
+config/price_reset_message = "Ihre exportierten Produkte auf Connect werden gelöscht, ihre Angebote werden dadurch ungültig.<br>Möchten Sie fortfahren?"
+config/price_reset_success = "Die exportierten Preise wurden erfolgreich zurückgesetzt. Es kann bis zu 10 Minuten dauern, bis diese Änderung Wirkung zeigt. Wenn die Änderung durchgeführt wurde, haben Sie beim Öffnen des Punktes "Produkte zu Connect" die Möglichkeit die zu exportierenden Preise neu zu setzen."
```
sbofirov: same here

Views/backend/connect/controller/main.js
```diff
@@ -1427,6 +1430,29 @@ Ext.define('Shopware.apps.Connect.controller.Main', {
...
+                var response = Ext.JSON.decode(result.responseText);
+                if (response.success) {
+                    me.createGrowlMessage(
+                        me.messages.priceResetLabel,
+                        me.messages.priceResetSuccess
+                    );
+                } else {
+                    me.createGrowlMessage(
+                        me.messages.priceResetLabel,
+                        response.message
```
sbofirov: This error is not translated. It comes from SN. Do we need to show some generic message and real reason to be in the logs?

Controllers/Backend/ConnectConfig.php
mkafadarov: Sorry about that I forgot to remove this method. I am not using it anywhere

PR #219, opened 2017-01-05
CON-3973 Import/Export article properties

mkafadarov:


Library/Shopware/Connect/SDK.php
```diff
@@ -60,7 +60,7 @@
     /**
      * Version constant
      */
-    const VERSION = '2.0.1';
+    const VERSION = '2.0.2';
```
sbofirov: please update SDK

Components/ProductToShop.php
```diff
@@ -435,6 +442,78 @@ public function insertOrUpdate(Product $product)
...
+        }
+
+        $article->setPropertyGroup($group);
+
+        $valueCollection = [];
+        $optionRepo = $this->manager->getRepository(PropertyOption::class);
+        $valueRepo = $this->manager->getRepository(PropertyValue::class);
+
+        foreach ($product->properties as $property) {
+
```
sbofirov: remove empty line

Components/ProductToShop.php
```diff
@@ -435,6 +442,78 @@ public function insertOrUpdate(Product $product)
...
+                ['property' => "groups.name", 'expression' => '=', 'value' => $property->groupName],
+            ];
+
+            $query = $groupRepo->getPropertyRelationQuery($filters, null, 1, 0);
+            $relation = $query->getOneOrNullResult();
+
+            if (!$relation) {
+                $group->addOption($option);
+                $this->manager->persist($group);
+                $this->manager->flush($group);
```
sbofirov: what about if remove this and use last flush without arguments at the end of this function?

Components/ProductQuery/LocalProductQuery.php
```diff
@@ -200,6 +201,8 @@ public function getConnectProduct($row)
         $row['url'] = $this->getUrlForProduct($row['sourceId']);
 		$row['priceRanges'] = $this->preparePriceRanges($row['detailId']);
 
+        $row['properties'] = $this->prepareProperties($row['localId']);
```
sbofirov: There isn't tests to confirm that product properties are really exported.

Components/ProductToShop.php
```diff
@@ -279,6 +283,9 @@ public function insertOrUpdate(Product $product)
             $model->setSupplier($supplier);
         }
 
+        //set product properties
+        $this->applyProductProperties($model, $product);
```
sbofirov: There isn't test to confirm that product properties are really imported in shopware article.

Subscribers/Article.php
```diff
@@ -1,11 +1,14 @@
 <?php
 
 namespace ShopwarePlugins\Connect\Subscribers;
+use ShopwarePlugins\Connect\Components\ErrorHandler;
 use Shopware\Connect\Struct\Change\FromShop\MakeMainVariant;
 use Shopware\Models\Customer\Group;
 use Shopware\Connect\Gateway;
 use Shopware\Components\Model\ModelManager;
 use ShopwarePlugins\Connect\Components\ConnectExport;
+use Shopware\Models\Article\Article as Product;
```
mkafadarov: Yeah, u r rigth, but the Subscriber class is also named Article so there is a conflict, I will rename it to ArticleModel

Components/ProductToShop.php
mkafadarov: If I remove the flush it will add the same option more times,  because the relation still not exists between group and option

Components/ProductToShop.php
mkafadarov: Done

Components/ProductQuery/LocalProductQuery.php
mkafadarov: Done

PR #224, opened 2017-01-16
CON-4126 Showed only static streams in stream export list

mkafadarov:


svenfinke:
[Bla Bla Bla Epic](https://trello.com/c/My8SLFGS/1-bla-bla-bla-epic)

PR #228, opened 2017-01-23
CON-4174 Fix multiple items with the same SKU

sbofirov:


Components/ConnectExport.php
```diff
@@ -165,34 +166,83 @@ public function export(array $ids, ProductStreamsAssignments $streamsAssignments
...
+        );
+
+        if (isset($params['export_status']) && !in_array($params['export_status'], $possibleValues)) {
+            throw new \InvalidArgumentException('Invalid export status');
+        }
+
+        if (isset($params['exported']) && !is_bool($params['exported'])) {
+            throw new \InvalidArgumentException('Parameter $exported must be boolean.');
+        }
+
```
mkafadarov: if $params are empty the dbal should NOT be executed

Subscribers/Lifecycle.php
```diff
@@ -215,32 +241,32 @@ public function onDeleteShop(\Enlight_Event_EventArgs $eventArgs)
...
-        if ($autoUpdate == 1) {
+        if ($this->autoUpdateProducts == 1 || $force === true) {
             $this->getConnectExport()->export(
                 array($attribute->getSourceId()), null, true
             );
-        } elseif ($autoUpdate == 2) {
+        } elseif ($this->autoUpdateProducts == 2) {
             $attribute->setCronUpdate(true);
             Shopware()->Models()->persist($attribute);
             Shopware()->Models()->flush();
```
mkafadarov: Please use the $this->manager

PR #229, opened 2017-01-24
CON-4098 - SDK change for the stock check change

StefanStefanoff:


Components/BasketHelper.php
```diff
@@ -570,6 +571,25 @@ public function getDefaultTemplateVariables()
      */
     public function getConnectTemplateVariables($connectMessages)
```
sbofirov: Please add type

PR #236, opened 2017-02-07
[CON-3775] Change translations

svenfinke:


sbofirov:
I think you have to rename ```config/import/overwrite_properties_helptext``` to something else, because if the plugin is already installed this snippet won't be changed. It's not released yet, but QA can fail because of that.

PR #237, opened 2017-02-09
CON-4047 Added conect order notify event

mkafadarov:


mkafadarov:
I have put the event between creation and check of the Connect Order, which is different from the ticket description, otherwise the modification will be overrided

PR #243, opened 2017-02-20
CON-4107 Connect Property groups cannot be assigned to products

StefanStefanoff:


Subscribers/Search.php
```diff
@@ -0,0 +1,100 @@
...
+        $request = $subject->Request();
+
+        switch ($request->getActionName()) {
+            case 'search':
+                $this->registerMyTemplateDir();
+                $this->registerMySnippets();
+
+                $entity = $request->getParam('entity', null);
+
+                switch ($entity) {
```
mkafadarov: You can use only one call of $this->markRecordsAsConnect
Like this:
switch ($entity) {
   case 'Shopware\Models\Property\Group':
   case 'Shopware\Models\Property\Option':
   case 'Shopware\Models\Property\Value':
     $subject->View()->data = $this->markRecordsAsConnect(
        $subject->View()->data,
        $entity
     );
     break;

Subscribers/Search.php
```diff
@@ -0,0 +1,100 @@
...
+                    case 'Shopware\Models\Property\Group':
```
mkafadarov: You can use only one call of $this->markRecordsAsConnect
Like this:
```
switch ($entity) {
   case 'Shopware\Models\Property\Group':
   case 'Shopware\Models\Property\Option':
   case 'Shopware\Models\Property\Value':
       $subject->View()->data = $this->markRecordsAsConnect(
           $subject->View()->data,
           $entity
       );
       break;
```

Subscribers/Search.php
```diff
@@ -0,0 +1,100 @@
...
+                break;
+        }
+    }
+
+    public function markRecordsAsConnect($data, $modelName)
+    {
+        $result = [];
+
+        foreach ($data as $row) {
+            $recordId = $row['id'];
```
sbofirov: we have to check if array key exists.

Subscribers/Search.php
sbofirov: I didn't find better solution. There isn't any event in Shopware_Controllers_Backend_Search::searchAction.
So please keep it in this way, just fix other comments.

Subscribers/Search.php
sbofirov: As @mkafadarov said please add use statements for Group, Option, Value and add Group::class instead of case 'Shopware\Models\Property\Group'. Same for other classes.

PR #244, opened 2017-02-20
CON-4303 Errors in the checkout

mkafadarov:


Subscribers/Checkout.php
```diff
@@ -352,6 +376,11 @@ public function reserveConnectProductsOnCheckoutFinish(\Enlight_Event_EventArgs
         try {
             /** @var $reservation \Shopware\Connect\Struct\Reservation */
             $reservation = $sdk->reserveProducts($order);
+
+            if (!$reservation->success) {
```
sbofirov: please change that to ```if (!$reservation || !$reservation->success) {```
Now it works, because shopware doesn't show all errors. $reservation could be = false
then normally php throws exception A PHP error occurred: Notice: Trying to get property of non-object, but
shopware handles it somehow.

PR #256, opened 2017-03-14
CON-4404 Limit to 10 images per product

sbofirov:


Components/ProductQuery/LocalProductQuery.php
```diff
@@ -207,14 +207,15 @@ public function getConnectProduct($row)
 
         $row['images'] = array();
         $mediaFiles = $this->localMediaService->getProductMedia($product, $this->productContext);
-
+        $mediaFiles = array_slice($mediaFiles, 0, 5);
```
mkafadarov: Can you please make 5 a constant, like IMAGE_LIMIT or what ever

PR #276, opened 2017-03-31
CON-4397 Added stream insert as SN stream from product insert

mkafadarov:


Views/backend/product_stream/view/selected_list/connect_product.js
```diff
@@ -1,6 +1,6 @@
 //{namespace name=backend/connect/view/main}
 //{block name="backend/product_stream/view/selected_list/product" append}
-Ext.define('Shopware.apps.ProductStream.controller.ConnectProduct', {
+Ext.define('Shopware.apps.ProductStream.view.selected_list.ConnectProduct', {
     override: 'Shopware.apps.ProductStream.view.selected_list.Product',
 
     snippets: {
```
sbofirov: please use connectSnippets

Views/backend/product_stream/view/selected_list/connect_windows.js
```diff
@@ -0,0 +1,29 @@
...
+
+        return productGrid
+    },
+
+    createSettingPanel: function() {
+        var me = this,
+            settingsPanel = this.callParent(arguments);
+
+        if (me.record.get('isConnect')) {
+            settingsPanel.nameField.disable();
```
sbofirov: is it possible to foreach all fields in settingsPanel and then disable() or just set disabled for the whole settingsPanel?

PR #285, opened 2017-04-12
CON-4491 - added notify events to manipulate products before export

dbeier:


Components/ProductQuery/LocalProductQuery.php
```diff
@@ -31,9 +32,14 @@ class LocalProductQuery extends BaseProductQuery
 
     protected $baseProductUrl;
 
-    /** @var \ShopwarePlugins\Connect\Components\Config $configComponent */
+    /** @var
+     * \ShopwarePlugins\Connect\Components\Config $configComponent
```
sbofirov: Please fix this dochint

Components/ProductQuery/LocalProductQuery.php
```diff
@@ -64,7 +87,8 @@ public function __construct(
         ProductTranslatorInterface $productTranslator,
         ContextServiceInterface $contextService,
         MediaService $storeFrontMediaService,
-        $mediaService = null
+        $mediaService = null,
+        Enlight_Event_EventManager $eventManager
```
sbofirov: I prefer arguments with default values at the end. So please swap ```$mediaService``` and ```$eventManager```

Components/ProductQuery/LocalProductQuery.php
```diff
@@ -85,6 +109,7 @@ public function __construct(
             null,
             ContextService::FALLBACK_CUSTOMER_GROUP
         );
+        $this->eventManager = $eventManager;
```
sbofirov: Move this up. Next to ```$this->contextService = $contextService;```

Components/ProductQuery/RemoteProductQuery.php
```diff
@@ -109,6 +124,15 @@ protected function getConnectProduct($row)
         $product = new Product(
             $row
         );
+
+        $this->eventManager->notify(
+            'Connect_Supplier_Get_Single_Product_Before',
```
sbofirov: Do we need to use same name in LocalProductQuery and RemoteProductQuery, Connect_Supplier_Get_Single_Product_Before? 
In LocalProductQuery we fetch product which will be export from our shop. So in this case the shop is supplier.
In RemoteProductQuery we fetch product which is imported from supplier. So in this case the shop is merchant.

Tests/Shopware/Connect/LocalProductQueryTest.php
```diff
@@ -116,7 +117,8 @@ public function getLocalProductQuery()
                 $this->productTranslator,
                 $this->contextService,
                 $this->localMediaService,
-                $this->mediaService
+                $this->mediaService,
+                Shopware()->Container()->get('events')
```
sbofirov: Then you need to swap this as well.

Tests/Shopware/Connect/ProductQueryTest.php
```diff
@@ -101,9 +101,14 @@ public function getProductQuery()
                     new MarketplaceGateway(Shopware()->Models()),
                     $this->productTranslator,
                     $this->contextService,
-                    $this->localMediaService
+                    $this->localMediaService,
+                    null,
```
sbofirov: Here you don't need to use null for mediaService argument if you change their places.

Components/ProductQuery/RemoteProductQuery.php
dbeier: @sbofirov so it makes only sense to implement it for the localProductGateway or not? In the RemoteProductQuery it currently don't make sense to implement an event here. Because for the insert of products from supplier to merchant there is already a event https://docs.enterprise.shopware.com/connect/connect_extend_plugin/ "Connect_Merchant_Saving_ArticleAttribute_Before"

sbofirov: Yes maybe we don't need for products from other suppliers. ProductFromShop means only product where you're supplier. Please remove it from RemoteProductQuery or change event name if we still need to support this.

PR #286, opened 2017-04-12
CON-4496 - add events to ProductToShop update method

sans0r:
this events are usefull if you need to work with the Shopware-Connect Plugin and make use of the custom attribute.
With this events you are able to:
* presave the connect-object, change the title (e.g. for translations or other modifications on product-data)
* do modifications on the article_detail (e.g. change categories or properties (translations))
* and do custom workflows if the connect-product is persisted

This events are necessary to make use of the custom attribute and some _not default_ actions.


Components/ProductToShop.php
```diff
@@ -186,12 +186,26 @@ public function commit()
...
+
         // todo@dn: Set dummy values and make product inactive
         if (empty($product->title) || empty($product->vendor)) {
             return;
         }
 
         $detail = $this->helper->getArticleDetailModelByProduct($product);
+        $detail = $this->eventManager->filter(
+            'Connect_Merchant_Get_Article_Detail_After',
+            $detail,
```
sbofirov: ```$detail``` can be null when the product doesn't exist in your DB. So I think this event must be moved on line 247, before ```if (!empty($product->sku)) {```. Then ```$detail``` object always exists. In some cases it will be completely empty, but exists. Otherwise in you classes you have to check ``` if (!$detail) { return; } else { //do something }.

@sans0r what do you think ?

dbeier:
@sbofirov @mkafadarov if you are ready to merge this PR, please ping me first. I need to update the documentation

Components/ProductToShop.php
sans0r: @sbofirov: I think this is exactly the purpose of this filter event. ;-)
In some cases you want to merge a non-connect product with an existing article in your shop. Then you can trigger the "merging" with this filter by running your custom "select" in the filter function and return the corresponding detail, but then I would do the `if` in the opponent way: 
`if($detail) {return $detail;} else { /* search article */ }`.
So it's intended that the null-`$detail` is going in the event.

It could be possible to only trigger this event if `$detail` is `null`, but i think this would block the usage of your example. So it would be nice to always trigger this event, but I agree the obligatory `if` on the beginning isn't nice...

sbofirov:
@dbeier please update documentation.

dbeier:
@all please don't merge, I first need to update the documentation, after that I will merge it

PR #287, opened 2017-04-18
CON-4498 Added product activation on stream assignment to category

mkafadarov:


Components/ProductStream/ProductStreamService.php
```diff
@@ -119,6 +119,26 @@ public function getStreamAssignments($streamId)
...
+    {
+        $attribute = $stream->getAttribute();
+
+        if (!$attribute) {
+            return false;
+        }
+
+        return (boolean) $attribute->getConnectIsRemote();
+    }
+
```
sbofirov: Missing dochint

Subscribers/Category.php
```diff
@@ -39,21 +48,47 @@ public function extendBackendCategory(\Enlight_Event_EventArgs $args)
...
+                    $subject->View()->data
+                );
+
+                $subject->View()->data = $this->expandCategories(
+                    $subject->View()->data,
+                    $expandedCategories
+                );
+                break;
+            case 'updateDetail':
+                $streamId = $request->getParam('streamId');
```
sbofirov: Maybe default value must be null for cases when streamId is not defined.

Components/ProductStream/ProductStreamRepository.php
```diff
@@ -301,6 +301,46 @@ public function removeMarkedStreamRelations()
...
+        }, $articleIds);
+
+        if (!$articleIds) {
+            return false;
+        }
+
+        $articleBuilder = $this->manager->getConnection()->createQueryBuilder();
+        $articleBuilder->update('s_articles', 'a')
+            ->set('a.active', ':active')
+            ->where('a.id IN (:articleIds)')
```
sbofirov: What about if $articleIds contains hundred or thousand ids ?

PR #296, opened 2017-05-11
CON-4523 - added index to sw_connect_change table to deny timeouts fo…

dbeier:
…r big data


Bootstrap/Setup.php
```diff
@@ -926,4 +926,12 @@ private function getMaxStateId()
 
         return 0;
     }
+
+    private function addIndexToChangeTable()
```
sbofirov: Please check Setup::createMyTables function. You don't need this. Just add your index in initial schema creation.

PR #300, opened 2017-05-18
CON-4533 - hide show splitted up shipping costs option

dbeier:


Views/backend/connect/view/config/general/form.js
```diff
@@ -416,6 +416,7 @@ Ext.define('Shopware.apps.Connect.view.config.general.Form', {
                     fieldLabel: me.snippets.separateShippingLabel,
                     labelWidth: me.defaults.labelWidth,
                     inputValue: 1,
+                    hidden : true,
```
dbeier: I just hide it for now because we are going to reimplement it later.

And I move the checkbox, it also makes sense to remove all the logic.. but like I said it makes no sense, because it's going to come back

PR #301, opened 2017-05-23
CON-4610 Filter remote categories in import tree by given shopId

sbofirov:


sbofirov:
It's bug fix for categories problem reported by lagome shop. They need to be filtered by parent and shopId.

Controllers/Backend/Import.php
```diff
@@ -73,10 +73,19 @@ public function getImportedProductCategoriesTreeAction()
                 break;
             case strpos($parent, '_stream_') > 0:
                 list($shopId, $stream) = explode('_stream_', $parent);
-                $categories = $this->getCategoryExtractor()->getRemoteCategoriesTreeByStream($stream, $shopId, $hideMapped);
+                $categories = $this->getCategoryExtractor()->getRemoteCategoriesTreeByStream($stream, $shopId, $hideMapped, $shopId);
                 break;
             default:
-                $categories = $this->getCategoryExtractor()->getRemoteCategoriesTree($parent, false, $hideMapped);
+                $node = $this->request->getParam('id');
+                preg_match('/^(shopId(\d+)~)(.*)$/', $node, $matches);
```
SimonBaeumer: imho regexes should be documented what they are doing. For me it is always pain to debug and to find out why a regex was placed in a particular position. What do you think?

Tests/Shopware/Connect/Component/CategoryExtractorTest.php
```diff
@@ -650,5 +650,51 @@ public function testGetRemoteCategoriesTreeByStream()
         $remoteCategories = $this->categoryExtractor->getRemoteCategoriesTreeByStream($stream, $shopId);
         $this->assertEquals($expected, $remoteCategories);
     }
+
+    /**
+     * Test concat shopId, categoryId and unique number
+     */
+    public function testConcatShopIdAndCategoryId()
```
SimonBaeumer: :+1:

Tests/Shopware/Connect/Component/CategoryExtractorTest.php
```diff
@@ -650,5 +650,51 @@ public function testGetRemoteCategoriesTreeByStream()
...
+    {
+        $shopId = 1;
+
+        $randomStringGenerator = $this->getMockBuilder('\\ShopwarePlugins\\Connect\\Components\\RandomStringGenerator')
```
SimonBaeumer: Shopware 5.2 has a requirement > php5.6, so you could use `RandomStringGenerator::class`. And you can use the short array syntax :)

Tests/Shopware/Connect/Component/CategoryExtractorTest.php
```diff
@@ -650,5 +650,51 @@ public function testGetRemoteCategoriesTreeByStream()
...
+            $randomStringGenerator
+        );
+
+        $argument = sprintf('shopId%s~%s', $shopId, '/Kleidung-unit/Hosen-unit/Hosentraeger-unit');
+        $randomStringGenerator->expects($this->once())
+            ->method('generate')
+            ->with($argument)
+            ->willReturn($argument . '1040');
+
+        $expected = array(
```
SimonBaeumer: I wrote the expected results in own variables too, but meanwhile I prefer to put it directly into the `assert*`-Calls. 

Like:
```
$this->assertEquals([
    [
             'id' => 'shopId1~/Kleidung-unit/Hosen-unit/Hosentraeger-unit1040',
             'categoryId' => '/Kleidung-unit/Hosen-unit/Hosentraeger-unit',
             'name' => 'Hosentraeger',
             'leaf' => true,
             'children' => array(),
              'cls' => "sc-tree-node",
             'expanded' => false
    ],
    $result
]);
```
Most of the time it is more obvious to find the expected result, further you dont need to jump around with the eyes to find the expected value if a test failed.
But the most important point is that it breaks the readability of the test because the setup of all values increases dramatically, especially with bigger arrays.
It's my preferred way, you can do it like this, just a suggestion if you like it :)

sbofirov: I agree with you, especially in cases where this variable is used only once.

PR #302, opened 2017-05-24
CON-4590 - moved events to manipulate/extend export functions to a be…

dbeier:
…tter place

### DIFF
--- a/Components/ConnectExport.php
+++ b/Components/ConnectExport.php
@@ -144,8 +144,6 @@ class ConnectExport
 
         $this->manager->beginTransaction();
 
-        //HIER
-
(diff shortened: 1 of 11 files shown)


Components/ConnectExport.php
```diff
@@ -94,10 +124,28 @@ public function getArticleDetailById($id)
...
+            'Connect_Supplier_Get_All_Products_Before',
+            [
+                'subject' => $this,
+                'products' => $connectItems
+            ]
+        );
+
         $this->manager->beginTransaction();
 
+        //HIER
```
svenfinke: I guess this comment should be removed.

PR #303, opened 2017-05-24
NTR - Add connect:endpoint:set to update the connect api endpoint via…

SimonBaeumer:
… CLI

### DIFF
--- a/Commands/ApiEndpointCommand.php
+++ b/Commands/ApiEndpointCommand.php
@@ -34,11 +34,6 @@ class ApiEndpointCommand extends ShopwareCommand
         $symfonyStyle = new SymfonyStyle($input, $output);
         $apiEndpoint = $input->getArgument('api-endpoint');
 
-        if (filter_var($apiEndpoint, FILTER_VALIDATE_URL) !== false) {
-            $symfonyStyle->error('Invalid url was passed: ' . $apiEndpoint);
-            exit(1);
-        }
-
         $host = parse_url($apiEndpoint, PHP_URL_HOST);
 
         $configWriter->save('connectDebugHost', $host);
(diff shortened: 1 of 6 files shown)


Commands/ApiEndpointCommand.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this
+            ->setName('connect:endpoint:set')
+            ->setDescription('Set the API endpoint to a custom location')
+            ->addArgument(
+                'api-endpoint',
+                InputArgument::REQUIRED,
+                'URL to of the API endpoint.'
+            )
+            ->addArgument(
```
sbofirov: Do we need shopId? Currently different subshops can't work separately with different environments of Connect/SEM

SimonBaeumer: You are right, I removed the shopId ;)

Commands/ApiEndpointCommand.php
```diff
@@ -0,0 +1,45 @@
...
+        $configWriter = $this->container->get('config_writer');
+        $symfonyStyle = new SymfonyStyle($input, $output);
+        $apiEndpoint = $input->getArgument('api-endpoint');
+
+        if (filter_var($apiEndpoint, FILTER_VALIDATE_URL) !== false) {
+            $symfonyStyle->error('Invalid url was passed: ' . $apiEndpoint);
+            exit(1);
+        }
+
+        $configWriter->save('connectDebugHost', $apiEndpoint);
```
sbofirov: Actually connectDebugHost must be ```connect.local```, ```connect.shopware.com``` or ```stage.connect.shopware.com```. So you need to strip schema.

PR #310, opened 2017-06-06
CON-4591 Improved property creation

sbofirov:
Didn't find a way how to simulate described bug, but I made some improvements and probably this problem won't occur in the future.

**First**: when option doesn't exist in merchant shop, PropertyValue doesn't exist as well. So we don't need to call DB in this case. 

**Second**: I've changed ```$valueRepo->findOneBy(['optionId' => $option->getId(), 'value' => $property->value]);``` to ```$valueRepo->findOneBy(['option' => $option, 'value' => $property->value])```, because there isn't setter and getter for optionId. It's just e property mapped to column "optionID". this column is also mapped to property "option" in this class. It has many to one connection. So I prefer to use ```PropertyValue::option``` instead of ```PropertyValue::optionID```.

### DIFF
--- a/Components/ProductToShop.php
+++ b/Components/ProductToShop.php
@@ -519,7 +519,7 @@ class ProductToShop implements ProductToShopBase
 
         foreach ($product->properties as $property) {
             $option = $optionRepo->findOneBy(['name' => $property->option]);
-
+            $optionExists = $option instanceof PropertyOption;
             if (!$option) {
                 $option = new PropertyOption();
                 $option->setName($property->option);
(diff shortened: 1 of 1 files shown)


Components/ProductToShop.php
```diff
@@ -511,15 +511,19 @@ private function applyProductProperties(ProductModel $article, Product $product)
...
         $article->setPropertyGroup($group);
 
-        $valueCollection = new ArrayCollection();
         $optionRepo = $this->manager->getRepository(PropertyOption::class);
         $valueRepo = $this->manager->getRepository(PropertyValue::class);
 
         foreach ($product->properties as $property) {
             $option = $optionRepo->findOneBy(['name' => $property->option]);
-
+            $optionExists = $option instanceof PropertyOption;
```
SimonBaeumer: Imho it is a code smell if there are too many temporary variables like this. I would prefer to write it directly into the `if` statement. Anyway, you get my approve it is just a little thing (over time it gets messy. files could save ~100 lines of code or more sometimes ;) )

sbofirov:
@SimonBaeumer what do you thing about this code then? Do we need to keep it while we can't reproduce the bug? Actually not possible to reproduce it on the environments of the customer.

SimonBaeumer:
@sbofirov Imho we should merge it until we can prove that it will fix the issue. If this code improves the proven bad behaviour than we can merge it :+1:

PR #316, opened 2017-06-20
CON-4645 - Fix event name

SimonBaeumer:


Tests/Shopware/Connect/Component/ProductFromShopTest.php
```diff
@@ -495,4 +379,41 @@ public function testOnPerformSync()
...
+
+        $this->assertStringStartsWith('SC-', $result);
+    }
+
+    /**
+     * @return ProductFromShop
+     */
+    private function createService()
+    {
+        return new ProductFromShop(
```
sbofirov: you can move that in setUp function. then don't need to call it in every test.

SimonBaeumer: Done ;)

PR #323, opened 2017-06-26
CON-4531 Verify if defaultMarketplace variable is empty

sbofirov:


Views/backend/connect/view/config/general/form.js
```diff
@@ -107,7 +107,7 @@ Ext.define('Shopware.apps.Connect.view.config.general.Form', {
         var me = this,
             elements = [];
 
-        if (defaultMarketplace == false) {
+        if (defaultMarketplace && defaultMarketplace == false) {
```
SimonBaeumer: Why not `!defaultMarketplace`?

sbofirov: this variable can be an empty string, '0' or '1'. Even when it's 0 or 1 variable is string, not int.
if it has value = '0', ```if (!defaultMarketplace)``` doesn't work as ```if (defaultMarketplace && defaultMarketplace == false) {```.

PR #328, opened 2017-07-04
CON-4690 - Add php-cs-fixer

SimonBaeumer:
Implemented CS-Fixer:
- Install git hooks via `./.githooks/install-hooks.sh`
- Added `php-cs-fixer` to the build
- Removed `5.3` build, it only increases the build duration (wasn't the best idea at that moment)
- While commiting with coding style errors you get notifications which files break our coding style
- You can find more helpful commands here: https://github.com/FriendsOfPHP/PHP-CS-Fixer


Models/Connect/ConfigRepository.php
```diff
@@ -1,12 +1,16 @@
...
+/**
+ * (c) shopware AG <info@shopware.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
 
 namespace   Shopware\CustomModels\Connect;
 
-use         \Shopware\Components\Model\ModelRepository;
+use         Shopware\Components\Model\ModelRepository;
```
sbofirov: is it correct ?

.php_cs.dist
```diff
@@ -0,0 +1,61 @@
+<?php
+$finder = PhpCsFixer\Finder::create()
+    ->exclude([ 'Library' ])
```
sbofirov: do we need to exclude Views folder? It contains smarty or JS files.

.php_cs.dist
```diff
@@ -0,0 +1,61 @@
...
+file that was distributed with this source code.
+EOF;
+
+return PhpCsFixer\Config::create()
+    ->setRules([
+        '@PSR2' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'blank_line_after_namespace' => true,
+        'blank_line_after_opening_tag' => true,
+        'blank_line_before_return' => true,
```
sbofirov: it's deprecated, please use ```blank_line_before_statement```

.php_cs.dist
SimonBaeumer: No, the cs fixer will only fetch php files, so we only need to exclude source code which does not belong to us (or the project)

Models/Connect/ConfigRepository.php
SimonBaeumer: Yes

Components/BasketHelper.php
```diff
@@ -645,7 +642,8 @@ public function findPercentagedIndividualVouchers($voucherCode)
             ->setParameter('voucherCode', $voucherCode);
 
 
-        return $builder->getQuery()->getResult();    }
+        return $builder->getQuery()->getResult();
```
SimonBaeumer: There is a statement which forces us to make a new line before every `return` statement but I cannot see any configuration for cases where a useless line was added... :/

.php_cs.dist
SimonBaeumer: I tried it and the php-cs-fixer crashes because he doesn't know that rule :/

.php_cs.dist
sbofirov: :+1:

.php_cs.dist
sbofirov: I saw that in their repository. Maybe we use different version.
It's not critical, so we can replace at some point in the future.

Components/BasketHelper.php
sbofirov: me too :(

PR #336, opened 2017-07-13
CON-4611 Check exported product on removeStreams

sbofirov:
Don't need to generate stream assignment changes for products which are not exported to Connect system anymore.
Connect will skip them, but don't need to replicate superfluous changes.

### DIFF
--- a/Components/ConnectExport.php
+++ b/Components/ConnectExport.php
@@ -385,6 +385,10 @@ class ConnectExport
     public function syncDeleteDetail(Detail $detail)
     {
         $attribute = $this->helper->getConnectAttributeByModel($detail);
+        // force fetching ConnectAttribute from DB
+        // if it was update via query builder
+        // changes are not visible, because of doctrine proxy cache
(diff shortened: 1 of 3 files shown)


sbofirov:
@SimonBaeumer actually removeStreamsAction must be refactored completely in the future, because it contains business logic. For now I decided to make a simple change, because my ticket is about generation of superfluous change which doesn't break synchronization. I think this can be part of complete refactoring in the future.

SimonBaeumer:
@sbofirov Can we create a ticket for that? And how critical do you think is that stream assignment?

sbofirov:
@SimonBaeumer yes I can. It's not critical, because if Connect receives stream assignment change for product which is removed it will skip it. The main problem is that Connect plugin generates not needed changes. Depending on the products count in removed stream could be possible to generate hundred  or even thousand wrong changes. I'll create a ticket about refactoring of this action, but what do you think, it must be part of refactoring epic or not ?

PR #341, opened 2017-07-20
CON-4731 Updated SDK

sbofirov:


Subscribers/SubscribeManager.php
```diff
@@ -32,6 +32,8 @@
      */
     public function __construct()
     {
+        parent::__construct();
```
SimonBaeumer: I think it is very critical to call the parten constructor here. If you take a look at `\Enlight_Class::__construct` I don't see any positive effect on it.

PR #347, opened 2017-07-31
CON-4761 Always inject stream name in category id

sbofirov:
If argument $stream is not null inject it in category id string.
It's needed to prepare correct Connect category tree for imported products.


sbofirov:
@SimonBaeumer currently is on possible to execute regression tests manually I think. Also they're not included in travis build. What do you think?

PR #352, opened 2017-08-07
CON-4751 Added index in s_plugin_connect_items

sbofirov:


Bootstrap/Update.php
```diff
@@ -395,4 +396,19 @@ private function removeDuplicatedMenuItems()
...
+     * Create an index by source_id and shop_id.
+     * It's most used combination to search for products in this table
+     */
+    private function addConnectItemsIndex()
+    {
+        if (version_compare($this->version, '1.1.1', '<=')) {
+            try {
+                $this->db->query('ALTER TABLE s_plugin_connect_items ADD INDEX source_id (source_id, shop_id)');
+            } catch (\Exception $e) {
+                // ignore it if exists
```
SimonBaeumer: Could you add a log entry? That would be very nice :+1:

sbofirov: @SimonBaeumer I added log message when creation of the index fails

sbofirov:
@SimonBaeumer I changed CategoryExtractorTest.Removed fifth argument ```$this->db```, because ```AutoCategoryResolver``` requires only 4 arguments.

PR #356, opened 2017-08-10
CON-4693 - Remove emotion template 

SimonBaeumer:


Subscribers/BaseSubscriber.php
```diff
@@ -112,20 +112,11 @@ public function registerMyTemplateDir()
...
-                $template = 'responsive';
-            } else {
-                $template = 'emotion';
-            }
+            $template = 'responsive';
         }
 
         $this->Application()->Template()->addTemplateDir(
-            $this->Path() . 'Views/' . $template, 'connect'
+            $this->Path() . "Views/$template/connect"
```
sbofirov: why you don't use ```'Views/responsive/connect'``` instead of ```$this->Path() . "Views/$template/connect"``` ?

SimonBaeumer: If the shop is not available (see in the if-statement), it loads the backend views. I use `"View/$template/connect" instead of concatenating it. Imho it imrpoves the readability a bit.

PR #364, opened 2017-08-28
CON-4807: remote categories can be mapped by multiple local categories

keulinho:


Bootstrap/Update.php
```diff
@@ -427,4 +428,42 @@ private function addConnectItemsIndex()
...
+                  `remote_category_id` int(11) NOT NULL,
+                  `local_category_id` int(11) unsigned NOT NULL,
+                  PRIMARY KEY (`remote_category_id`, `local_category_id`),
+                  CONSTRAINT s_plugin_connect_remote_categories_fk_remote_category_id FOREIGN KEY (remote_category_id) REFERENCES s_plugin_connect_categories (id) ON DELETE CASCADE,
+                  CONSTRAINT s_plugin_connect_remote_categories_fk_local_category_id FOREIGN KEY (local_category_id) REFERENCES s_categories (id) ON DELETE CASCADE
+                  ) ENGINE = InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;'
+                );
+                $result = $this->db->query('SELECT pcc.id, pcc.local_category_id
+                                FROM s_plugin_connect_categories pcc
+                                WHERE pcc.local_category_id is not null');
```
sbofirov: please use capital letters for MySQL keywords reserved words. In this case ```IS NOT NULL```

Bootstrap/Update.php
```diff
@@ -427,4 +428,42 @@ private function addConnectItemsIndex()
...
+
+                while ($row=$result->fetch()) {
```
sbofirov: Missing one space between ```=```
Maybe you need to run ```bash .githooks/install-hooks.sh``` locally. After that before every commit php cs fixer will scan your code and notify you have some places where code doesn't have correct code style.

Models/Connect/RemoteCategory.php
```diff
@@ -97,34 +97,36 @@ public function setLabel($label)
...
-        $this->localCategoryId = $id;
+        return $this->localCategories->toArray();
     }
 
     /**
-     * @return \Shopware\Models\Category\Category
+     * @param \Shopware\Models\Category\Category $localCategory
      */
-    public function getLocalCategory()
+    public function addLocalCategory($localCategory)
```
sbofirov: Please use typehint. It must be ```public function addLocalCategory(Category $localCategory)``` where Category is Shopware\Models\Category\Category.

Models/Connect/RemoteCategory.php
```diff
@@ -97,34 +97,36 @@ public function setLabel($label)
...
+        if (!$this->localCategories->contains($localCategory)) {
+            $this->localCategories->add($localCategory);
+        }
     }
 
     /**
      * @param \Shopware\Models\Category\Category $localCategory
      */
-    public function setLocalCategory($localCategory)
+    public function removeLocalCategory($localCategory)
```
sbofirov: same here

Bootstrap/Update.php
keulinho: the cs fixer ran but i wil fix this

PR #370, opened 2017-08-30
[Con 4768] delete the merchants Category mapping for removed categories  

dbeier:


Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -103,7 +115,31 @@ public function storeRemoteCategories(array $categories, $articleId)
...
             $this->manager->persist($productToCategory);
         }
+    }
 
-        $this->manager->flush();
+    /**
+     * @param \Shopware\CustomModels\Connect\RemoteCategory[] $assignedCategories
+     * @param $articleId
+     */
+    private function removeProductsFromRemoteCategory($assignedCategories, $articleId)
```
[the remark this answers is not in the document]
dbeier: " @param \Shopware\CustomModels\Connect\RemoteCategory[] $assignedCategories" or do you mean something else

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -103,7 +115,31 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+    {
+        $currentProductCategories = $this->productToRemoteCategoryRepository->getArticleRemoteCategories($articleId);
+
+        /** @var ProductToRemoteCategory $currentProductCategory */
+        foreach ($currentProductCategories as $currentProductCategory) {
```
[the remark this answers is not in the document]
dbeier: @sbofirov sure that this one will work?

if (!in_array($currentProductCategory, $assignedCategoryIds) {
$this->manager->remove($currentProductCategory);
}

Because the Array is an Array of Objects... and the assignedCategoryIds are integer values... is this working? I never tried....

Components/CategoryResolver/DefaultCategoryResolver.php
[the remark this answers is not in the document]
sbofirov: ```private function removeProductsFromRemoteCategory(array $assignedCategories, $articleId)```

Components/CategoryResolver/DefaultCategoryResolver.php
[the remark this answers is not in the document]
sbofirov: Ahh yes, good point. must be ```$currentProductCategory->getConnectCategoryId()``` instead of ```$currentProductCategory```

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -103,7 +115,27 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+        $assignedCategoryIds = array_map(function (RemoteCategory $assignedCategory) {
+            $assignedCategory->getId();
+        }, $assignedCategories);
+
+        /** @var ProductToRemoteCategory $currentProductCategory */
+        foreach ($currentProductCategories as $currentProductCategory) {
+            if (!in_array($currentProductCategory->getConnectCategoryId(), $assignedCategoryIds)) {
+                //DELETE PRODUCT ASSIGNMENT
+                $this->manager->remove($currentProductCategory);
+                $this->manager->flush();
```
sbofirov: why flush is in foreach ?

Tests/Unit/Components/CategoryResolver/DefaultCategoryResolverTest.php
```diff
@@ -0,0 +1,49 @@
...
+            $this->createMock(ProductToRemoteCategoryRepository::class)
+        );
+    }
+
+    public function test_store_remote_categories_success()
+    {
+        $articleId = 4;
+        $categories = ['Category 1'];
+
+        $this->defaultCategoryResolver->storeRemoteCategories($categories, $articleId);
```
sbofirov: Actually you need to verify which mock was called, which arguments were used ...
Right now this test only make a call and doesn't verify anything.

dbeier: A I thought it's enough to ensure that the test is running.. there are no callbacks. But I will add something like that fn xyz is called x times

Tests/Unit/Components/CategoryResolver/DefaultCategoryResolverTest.php
```diff
@@ -0,0 +1,61 @@
...
+            $this->manager,
+            $this->remoteCategoryRepository,
+            $this->createMock(ProductToRemoteCategoryRepository::class)
+        );
+    }
+
+    public function test_store_remote_categories_success()
+    {
+        $articleId = 4;
+        $categories = ['Category 1'];
```
sbofirov: Would be better to use ```$categories = ['category1' => 'Category 1'];``` instead of ```$categories = ['Category 1'];```, because we use this structure in Product::categories.
Btw your test handles only 1 case. 

@svenfinke what do you think ?

dbeier: I will add some fails etc to the tests and fix that category thing

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -100,7 +112,26 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+    }
 
-        $this->manager->flush();
+    /**
+     * @param \Shopware\CustomModels\Connect\RemoteCategory[] $assignedCategories
+     * @param $articleId
+     */
+    private function removeProductsFromRemoteCategory(array $assignedCategories, $articleId)
+    {
+        $currentProductCategories = $this->productToRemoteCategoryRepository->getArticleRemoteCategories($articleId);
```
JanPietrzyk: How about using ids only here?

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -100,7 +112,26 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+
+        $assignedCategoryIds = array_map(function (RemoteCategory $assignedCategory) {
+            $assignedCategory->getId();
+        }, $assignedCategories);
+
+        /** @var ProductToRemoteCategory $currentProductCategory */
+        foreach ($currentProductCategories as $currentProductCategory) {
+            if (!in_array($currentProductCategory->getConnectCategoryId(), $assignedCategoryIds)) {
+                //DELETE PRODUCT ASSIGNMENT
+                $this->manager->remove($currentProductCategory);
```
JanPietrzyk: would be much less load if you just use delete DQL/SQL here - maybe as a repository method?

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -85,6 +85,18 @@ public function storeRemoteCategories(array $categories, $articleId)
...
+
+    /**
+     * @param RemoteCategory[] $remoteCategories
+     * @param $articleId
+     */
+    private function addProductToRemoteCategory($remoteCategories, $articleId)
+    {
         /** @var $remoteCategory \Shopware\CustomModels\Connect\RemoteCategory */
         foreach ($remoteCategories as $remoteCategory) {
             $productToCategory = $this->productToRemoteCategoryRepository->findOneBy([
```
JanPietrzyk: Is it possible to extract this fetch statement, to be outside of the `foreach`?

PR #371, opened 2017-08-31
[CON-4815] Replace Doctrine with more performant services

svenfinke:


Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -124,6 +124,28 @@ public function convertTreeToEntities(array $node, Category $parent = null, $cat
     }
 
     /**
+     * Loop through category tree and fetch ids
+     *
+     * @param array $node
+     * @param array $categories
+     * @return array
+     */
+    public function convertTreeToKeys(array $node, $categories = [])
```
sbofirov: This could be private function in ImportService, otherwise we have to add it in CategoryResolver interface and implement it in both CategoryResolvers.

Controllers/Backend/Import.php
```diff
@@ -237,7 +237,29 @@ public function assignRemoteToLocalCategoryAction()
         }
 
         try {
-            $this->getImportService()->importRemoteCategory($localCategoryId, $remoteCategoryKey, $remoteCategoryLabel);
+            $categories = $this->getImportService()->createCategoriesFromRemoteCategoires($localCategoryId, $remoteCategoryKey, $remoteCategoryLabel);
```
sbofirov: In the past ```$this->getImportService()->importRemoteCategory($localCategoryId, $remoteCategoryKey, $remoteCategoryLabel);``` was called to create necessary entities and store them into DB.

Now ```createCategoriesFromRemoteCategoires``` will generate categoryTree and extract category keys  ```return $this->autoCategoryResolver->convertTreeToKeys($remoteCategoryNodes, $localCategory);```. Then this action returns response success true, but categories are not imported.

Controllers/Backend/Import.php
```diff
@@ -237,7 +237,29 @@ public function assignRemoteToLocalCategoryAction()
...
+            return;
+        }
+
+        $this->View()->assign([
+            'success' => true,
+            'categories' => json_encode($categories)
+        ]);
+    }
+
+    public function assignCategoriesToProductsAction()
```
sbofirov: Where it's used ?

Components/ImportService.php
```diff
@@ -240,29 +256,7 @@ public function importRemoteCategory($localCategoryId, $remoteCategoryKey, $remo
...
-        foreach ($categories as $category) {
-            $articleIds = $this->productToRemoteCategoryRepository->findArticleIdsByRemoteCategory($category['categoryKey']);
-
-            while ($currentIdBatch = array_splice($articleIds, 0, 10)) {
-                $articles = $this->articleRepository->findBy(['id' => $currentIdBatch]);
-                /** @var \Shopware\Models\Article\Article $article */
-                foreach ($articles as $article) {
-                    /** @var \Shopware\Models\Category\Category $categoryModel */
-                    $categoryModel = $category['model'];
-                    if ($article->getCategories()->contains($categoryModel->getParent())) {
```
sbofirov: We have to keep this logic as well, because this is needed when some customers assign their products to categories which are not leaf. For example Mr.Strauss need this.

I'm not sure but now you assign all categories. That means if product comes with
following categories: 
How it looks like in product struct
```
$product->categories = [
'category1' => 'Category 1',
'category1/category2' => 'Category2',
'category1/category2/category3' => 'Category3'
]
```
How it looks like as a tree
Category 1 
----- Category 2 
----- ----- Category 3

After product import it will be assigned to Category1, Category2, Category3 and these assignments will be visible in shopware backend. Have you tested this?

Subscribers/ServiceContainer.php
```diff
@@ -148,7 +148,9 @@ public function onImportService()
                 new PDO($this->db->getConnection()),
                 new RandomStringGenerator(),
                 $this->db
-            )
+            ),
+            $this->container->get('CategoryDenormalization'),
+            $this->container->get('shopware_attribute.data_persister')
```
sbofirov: Tests are failing in Travis. You need to  fix them.
```
PHP Catchable fatal error:  Argument 9 passed to ShopwarePlugins\Connect\Components\ImportService::__construct() must be an instance of Shopware\Components\Model\CategoryDenormalization, none given, called in /tmp/shopware/engine/Shopware/Plugins/Local/Backend/SwagConnect/Tests/Legacy/Shopware/Connect/Component/ImportServiceTest.php on line 79 and defined in /tmp/shopware/engine/Shopware/Plugins/Local/Backend/SwagConnect/Components/ImportService.php on line 82
	PHP Stack trace:
	PHP   1. {main}() /tmp/shopware/vendor/phpunit/phpunit/phpunit:0
```

PR #374, opened 2017-09-04
NTR Fix tests

sbofirov:


.psh.yml.dist
```diff
@@ -1,3 +1,29 @@
+header: |
```
keulinho: please rename to .psh.yml
it prevents errors like this one
because psh only reads the psh.yml.dist if theres no psh.yml

PR #375, opened 2017-09-08
[Con 4814] Dont load products if a stream is selected in import window

Christian-Rades:
Due to timeouts when selecting a stream in the import window, the products wont be loaded if a stream is selected


Controllers/Backend/Import.php
```diff
@@ -96,9 +96,19 @@ public function loadArticlesByRemoteCategoryAction()
...
-            $stream = $stream[1];
-            $category = null;
+
+            $snippets = Shopware()->Snippets()->getNamespace('backend/connect/view/main');
+
+            $this->View()->assign([
+                'success' => false,
+                'message' => $snippets->get(
+                    'import/message/dont_select_streams',
+                    'Please select a category instead of a stream',
```
sbofirov: Please use same english message as in main.ini or copy this to main.ini file.

Controllers/Backend/Import.php
```diff
@@ -96,9 +96,19 @@ public function loadArticlesByRemoteCategoryAction()
         $stream = $this->request->getParam('stream', null);
 
         if (strpos($category, '_stream_') > 0) {
-            $stream = explode('_stream_', $category);
-            $stream = $stream[1];
-            $category = null;
+
+            $snippets = Shopware()->Snippets()->getNamespace('backend/connect/view/main');
```
sbofirov: It's exactly the same, but please use ```$this->get('snippets')``` instead of ```Shopware()->Snippets()```

PR #376, opened 2017-09-08
CON-4671 Fixed export big streams

sbofirov:


Components/ProductStream/ProductStreamRepository.php
```diff
@@ -402,13 +402,20 @@ public function fetchConnectStreamIds()
     public function countProductsInStaticStream($streamId)
```
svenfinke: Can you add the valid types of $steamId to the Dochint? (int | array)

PR #377, opened 2017-09-11
CON-4857: fixed autocategory assignment in ProductToShop

keulinho:


sbofirov:
@keulinho can you try to implement some function which can be used in both places (ProductToShop & ImportService)?

keulinho:
implemented some ideas from svens ticket: https://jira.shopware.com/browse/CON-4815 from https://github.com/shopware/SwagConnect/compare/CON-4815-old
and added some tests

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -73,51 +73,47 @@ public function resolve(array $categories)
...
     }
 
     /**
-     * Loop categories tree recursive and
-     * create same structure with entities
+     * Loop through category tree and fetch ids
      *
      * @param array $node
-     * @param null Category $parent
+     * @param string $parentId
```
sbofirov: I guess it must be int

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -73,51 +73,47 @@ public function resolve(array $categories)
...
+     * @param bool $returnOnlyLeafs
      * @param array $categories
      * @return array
      */
-    public function convertTreeToEntities(array $node, Category $parent = null, $categories = [])
+    public function convertTreeToKeys(array $node, $parentId, $returnOnlyLeafs = true, $categories = [])
```
sbofirov: If every public function should be defined in CategoryResolver abstract class as abstract function or implemented there and available for every different implementation of CategoryResolver, otherwise it must be private, but we need it in ImportService.

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -227,4 +227,23 @@ private function getCategoryData($name)
                 ],
         ];
     }
+
+    /**
+     * @param array $category
+     * @param string $parentId
```
sbofirov: same here, it must be int.

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -227,4 +227,23 @@ private function getCategoryData($name)
...
+     * @return Category
```
sbofirov: it returns int

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -126,13 +122,17 @@ public function convertTreeToEntities(array $node, Category $parent = null, $cat
 
     /**
      * @param array $category
-     * @param Category $parent
+     * @param string $parentId
```
sbofirov: int

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -126,13 +122,17 @@ public function convertTreeToEntities(array $node, Category $parent = null, $cat
...
      * @return Category
      */
-    public function convertNodeToEntity(array $category, Category $parent)
+    public function convertNodeToEntity(array $category, $parentId)
     {
         $categoryModel = new Category();
         $categoryModel->fromArray($this->getCategoryData($category['name']));
+
+        $parent = $this->categoryRepository->findOneBy([
+            'id' => $parentId
```
sbofirov: you can cast it to int. Useful when someone uses this function with string values. Doctrine won't cast automatically.
I had problems with varchar column in db and int values. then index doesn't work.

sbofirov: If parent doesn't exist then it will throw an exception "Call getId on null", but I think it's fine

Components/ImportService.php
```diff
@@ -240,27 +256,35 @@ public function importRemoteCategory($localCategoryId, $remoteCategoryKey, $remo
...
+                    [$articleId,  $category['categoryKey']]
+                );
+                $this->manager->getConnection()->executeQuery(
+                    'DELETE FROM `s_articles_categories` WHERE `articleID` = :articleID AND `categoryID` = :categoryID',
+                    [
+                        ':articleID' => $articleId,
+                        ':categoryID' => $category['parentId']
+                    ]
+                );
+                $detailId = $this->manager->getConnection()->fetchColumn(
```
sbofirov: Do we need to fetch detailId? I think we can use only articleId in query below.

Components/ImportService.php
```diff
@@ -240,27 +256,35 @@ public function importRemoteCategory($localCategoryId, $remoteCategoryKey, $remo
...
+                    'SELECT id FROM `s_articles_details` WHERE `articleID` = :articleID',
+                    ['articleID' => $articleId]
+                );
+                $this->manager->getConnection()->executeQuery(
+                    'INSERT  INTO `s_articles_attributes` (`articleID`, `articledetailsID`, `connect_mapped_category`) 
+                        VALUES (?, ?, 1)
+                        ON DUPLICATE KEY UPDATE `connect_mapped_category` = 1
+                    ',
+                    [$articleId,  $detailId]
```
sbofirov: will work only with articleId I think

Components/CategoryResolver/AutoCategoryResolver.php
```diff
@@ -73,51 +73,47 @@ public function resolve(array $categories)
...
                 'parentId' => 1,
             ]);
+            // if connectTree has a Subtree starting with Spanish but MerchantShop has no mainCategory Spanish
+            // the categories below Spanish won't be created
+            if ($mainCategory == null) {
+                continue;
+            }
 
-            $remoteCategories = array_merge($remoteCategories, $this->convertTreeToEntities($node['children'], $mainCategory));
+            $remoteCategories = $this->convertTreeToKeys($node['children'], $mainCategory->getId());
```
sbofirov: AutoCategoryResolver still returns only leaf categories.
Won't be possible to use it and assign categories to categories which are not leaf.

keulinho: convert tree to keys has a bool parameter if only leafs will be returned 
per default it is set to true
the import service calls the convertToKeys method directly with this parameter set to false
the resolve function is just called in productToShop and there it is right to return just leafs

Components/CategoryResolver/AutoCategoryResolver.php
keulinho: i just move it to the abstract class because there is just this one implementation

Components/ProductToShop.php
```diff
@@ -433,6 +435,20 @@ public function insertOrUpdate(Product $product)
 
         $this->manager->flush();
 
+        //article has to be flushed
+        $categories = $this->categoryResolver->resolve($product->categories);
+        $hasMappedCategory = count($categories) > 0;
+        $detailAttribute->setConnectMappedCategory($hasMappedCategory);
```
sbofirov: there isn't persist of ```detailAttribute```. Also flush was before this change.
If you see on some line below we call ```$this->manager->clear();```.
So this change won't be stored in db.

Tests/Legacy/Shopware/Connect/Component/CategoryResolver/AutoCategoryResolverTest.php
```diff
@@ -78,33 +78,30 @@ public function testResolve()
...
+            '/deutsch' => 'Deutsch',
         ];
 
         $categoryModels = $this->categoryResolver->resolve($categories);
 
-        //Spanish category must not be created
-        $this->assertNull($this->categoryRepo->findOneByName('Spanish'));
         $this->assertEquals(
             $defaultCategoryId,
             Shopware()->Db()->fetchOne('SELECT parent FROM s_categories WHERE description = ?', ['Adidas'])
```
sbofirov: Please ask Charly about that. In the past we just imported Adidas category into Deutsch

Tests/Legacy/Shopware/Connect/Component/ImportServiceTest.php
```diff
@@ -190,4 +198,34 @@ public function testFindRemoteArticleIdsByCategoryId()
...
     }
+
+    public function testImportRemoteCategory()
+    {
+        $this->manager->getConnection()->setTransactionIsolation(\Doctrine\DBAL\Connection::TRANSACTION_READ_UNCOMMITTED);
+        $this->manager->getConnection()->beginTransaction();
+        $localCategory = $this->categoryRepository->find(14);
+        /** @var \Shopware\CustomModels\Connect\RemoteCategory $remoteCategory */
+        $remoteCategory = $this->remoteCategoryRepository->findOneBy(['categoryKey' => '/deutsch/bücher']);
+
```
sbofirov: not needed empty line

Tests/Legacy/Shopware/Connect/Component/ImportServiceTest.php
```diff
@@ -190,4 +198,34 @@ public function testFindRemoteArticleIdsByCategoryId()
...
+
+        $this->importService->importRemoteCategory($localCategory->getId(), $remoteCategory->getCategoryKey(), $remoteCategory->getLabel());
+
+        $createdLocalCategory = $this->categoryRepository->findOneBy([
+            'name' => $remoteCategory->getLabel(),
+            'parent' => $localCategory->getId()
+            ]);
+
+        $this->assertTrue($createdLocalCategory!=null);
```
sbofirov: you can use ```$this->assertNotNull``` instead of this or better ```$this->assertInstanceOf()```

Tests/Legacy/Shopware/Connect/ProductToShopTest.php
```diff
@@ -524,12 +529,27 @@ public function testInsertArticleAndAutomaticallyCreateCategories()
             $childCategoryModel->getParent()->getName()
         );
 
+        $categoryRepository = Shopware()->Models()->getRepository('Shopware\Models\Category\Category');
```
sbofirov: please add use statement and call it with ```Category::class```

Tests/Legacy/Shopware/Connect/ProductToShopTest.php
```diff
@@ -524,12 +529,27 @@ public function testInsertArticleAndAutomaticallyCreateCategories()
...
+        /** @var \Shopware\Models\Category\Category $childCategoryModel */
+        $childCategoryModel2 = $categoryRepository->findOneBy(['name' => $childCategory2]);
+        $this->assertInstanceOf('Shopware\Models\Category\Category', $childCategoryModel2);
```
sbofirov: Same here

Tests/Legacy/Shopware/Connect/ProductToShopTest.php
```diff
@@ -524,12 +529,27 @@ public function testInsertArticleAndAutomaticallyCreateCategories()
...
+
         /** @var Article $article */
         $article = $this->modelManager->getRepository(Article::class)->findOneByName($product->title);
 
         $assignCategories = $article->getCategories();
-        $this->assertEquals(1, count($assignCategories));
-        $this->assertEquals($childCategory, $assignCategories[0]->getName());
+        $this->assertEquals(2, count($assignCategories));
+        //this category is assigned to the product when it is created
+        $this->assertEquals('Bücher', $assignCategories[0]->getName());
```
sbofirov: why? Probably, because Bücher was mapped to local category in previous test ?

Components/ImportService.php
keulinho: no the tests fail then 
serviceContainerTest::testImportCategory

Tests/Legacy/Shopware/Connect/ProductToShopTest.php
keulinho: no it's hard coded in $this->getProduct()
i just overwrite the categories instead of merge them

Tests/Legacy/Shopware/Connect/Component/CategoryResolver/AutoCategoryResolverTest.php
keulinho: in the past Adidas would get parent null and because of that it wasn't visible
it doesn't make sense to import it in the default category because the category names will be in Spanish and that would be pretty confusing

Components/ConnectFactory.php
```diff
@@ -201,7 +201,8 @@ public function createSDK()
             new DefaultCategoryResolver(
                 $manager,
                 $manager->getRepository('Shopware\CustomModels\Connect\RemoteCategory'),
-                $manager->getRepository('Shopware\CustomModels\Connect\ProductToRemoteCategory')
+                $manager->getRepository('Shopware\CustomModels\Connect\ProductToRemoteCategory'),
+                $manager->getRepository('Shopware\Models\Category\Category')
```
sbofirov: Please replace Shopware\CustomModels\Connect\ProductToRemoteCategory and Shopware\Models\Category\Category with ProductToRemoteCategory::class and Category::class

PR #382, opened 2017-09-15
ToShop performance improvements: remove entitymanager from AutoCatego…

keulinho:
…ryResolver and remove unnessecary sort in generateTree


keulinho:
Performance Improvement locally with 20 Products and each with 200 categories assigned:
from 90s -> to 23s

PR #390, opened 2017-09-25
CON-4880: fixed problem in store remote categories

keulinho:


Components/CategoryResolver.php
```diff
@@ -90,8 +90,8 @@ public function storeRemoteCategories(array $categories, $articleId)
 
         $this->manager->flush();
 
-        $this->addProductToRemoteCategory($remoteCategories, $articleId);
         $this->removeProductsFromRemoteCategory($remoteCategories, $articleId);
+        $this->addProductToRemoteCategory($remoteCategories, $articleId);
```
sbofirov: what is the benefit to call ```addProductToRemoteCategory``` after ```addProductToRemoteCategory```?
Less iterations in removeProductsFromRemoteCategory?

Components/CategoryResolver.php
```diff
@@ -129,10 +129,7 @@ private function removeProductsFromRemoteCategory(array $assignedCategories, $ar
         /** @var int $currentProductCategoryId */
         foreach ($currentProductCategoryIds as $currentProductCategoryId) {
             if (!in_array($currentProductCategoryId, $assignedCategoryIds)) {
-                $this->productToRemoteCategoryRepository->deleteByConnectCategoryId($currentProductCategoryId);
-                if ($this->productToRemoteCategoryRepository->getProductCountByCategoryId($currentProductCategoryId) > 0) {
-                    $this->remoteCategoryRepository->deleteById($currentProductCategoryId);
-                }
+                $this->productToRemoteCategoryRepository->deleteByConnectCategoryId($currentProductCategoryId, $articleId);
```
sbofirov: Why you removed the check where cont if there is at least 1 product in remote category? Otherwise remove it completely.

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+    private $manager;
+    private $config;
+    private $categoryRepo;
+    private $categoryResolver;
+
+    public function setUp()
+    {
+        $this->manager = Shopware()->Models();
+        $this->config = ConfigFactory::getConfigInstance();
+        $this->categoryRepo = $this->manager->getRepository('Shopware\Models\Category\Category');
```
sbofirov: Please use ```Category:class```

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $this->categoryResolver = new AutoCategoryResolver(
+            $this->manager,
+            $this->categoryRepo,
+            $this->manager->getRepository('Shopware\CustomModels\Connect\RemoteCategory'),
```
sbofirov: Same here ```RemoteCategory::class```

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+            $this->config,
+            $this->manager->getRepository(ProductToRemoteCategory::class)
+        );
+    }
+
+    public function testStoreRemoteCategories()
+    {
+        $this->manager->getConnection()->executeQuery('DELETE FROM s_plugin_connect_categories');
+        $this->manager->getConnection()->executeQuery('DELETE FROM s_plugin_connect_product_to_categories');
```
sbofirov: why you need both delete lines ? Integration tests already work with transactions, so it should be a problem only to insert records.

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $this->manager->getConnection()->executeQuery('INSERT INTO s_plugin_connect_categories (category_key, label) VALUES ("/deutsch/test1", "Test 1")');
+        $this->manager->getConnection()->executeQuery('INSERT INTO s_plugin_connect_product_to_categories (connect_category_id, articleID) VALUES (LAST_INSERT_ID(), 3)');
+        $this->manager->getConnection()->executeQuery('INSERT INTO s_plugin_connect_categories (category_key, label) VALUES ("/deutsch/test2", "Test 2")');
+        $this->manager->getConnection()->executeQuery('INSERT INTO s_plugin_connect_product_to_categories (connect_category_id, articleID) VALUES (LAST_INSERT_ID(), 3)');
+
+        $germanID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch"')->fetchColumn();
+        $test1ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test1"')->fetchColumn();
+        $test2ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test2"')->fetchColumn();
+
```
sbofirov: not needed empty line

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $categories = [
+            '/deutsch' => 'Deutsch',
+            '/deutsch/test1' => 'Test 1',
+            '/deutsch/test3' => 'Test 3',
+            '/deutsch/test3/test31' => 'Test 3.1'
+        ];
+
+        $this->categoryResolver->storeRemoteCategories($categories, 3);
+
+        $actualGermanID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch"')->fetchColumn();
```
sbofirov: please use camelCase. ```actualGermanId``` instead of ```actualGermanID```

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $actualTest1ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test1"')->fetchColumn();
+        $actualTest2ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test2"')->fetchColumn();
+
+        //Assert that old remote categories aren't changed
+        $this->assertEquals($germanID, $actualGermanID);
+        $this->assertEquals($test1ID, $actualTest1ID);
```
sbofirov: same here, camelCase

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $this->assertEquals($test2ID, $actualTest2ID);
+
+        $test3ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test3"')->fetchColumn();
+        $test31ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test3/test31"')->fetchColumn();
+
+        //Assert that new Categories are created
+        $this->assertInternalType('string', $test3ID);
+        $this->assertInternalType('string', $test31ID);
```
sbofirov: why we need to check that returned ids are strings?

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $actualTest1ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test1"')->fetchColumn();
+        $actualTest2ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test2"')->fetchColumn();
+
+        //Assert that old remote categories aren't changed
+        $this->assertEquals($germanID, $actualGermanID);
+        $this->assertEquals($test1ID, $actualTest1ID);
+        $this->assertEquals($test2ID, $actualTest2ID);
+
+        $test3ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test3"')->fetchColumn();
+        $test31ID = $this->manager->getConnection()->executeQuery('SELECT id FROM s_plugin_connect_categories WHERE category_key = "/deutsch/test3/test31"')->fetchColumn();
```
sbofirov: camelCase for both lines

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        //Assert that new Categories are created
+        $this->assertInternalType('string', $test3ID);
+        $this->assertInternalType('string', $test31ID);
+
+        $productToCategoryID = $this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = 3 AND connect_category_id = $test1ID")->fetchColumn();
```
sbofirov: same here

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        //Assert that old category is still assigned
+        $this->assertInternalType('string', $productToCategoryID);
```
sbofirov: same here

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $productToCategoryID = $this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = 3 AND connect_category_id = $test2ID")->fetchColumn();
```
sbofirov: same here

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        //Assert that removed category is not assigned
+        $this->assertEquals(false, $productToCategoryID);
+
+        $productToCategoryID = $this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = 3 AND connect_category_id = $test3ID")->fetchColumn();
```
sbofirov: better set where parameters as a second argument array of ```exceuteQuery```
Instead ```$this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = 3 AND connect_category_id = $test3ID")->fetchColumn();``` you can call ```$this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = ? AND connect_category_id = ?", [3, $test3ID])->fetchColumn();```

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        //Assert that new, not-leaf category is assigned
+        //This is necessary that ext.js find all products in not-leaf categories in import window
+        $this->assertInternalType('string', $productToCategoryID);
```
sbofirov: I think it looks better to check that ```$productToCategoryID``` is greater than zero.

Tests/Integration/Components/CategoryResolverTest.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $productToCategoryID = $this->manager->getConnection()->executeQuery("SELECT id FROM s_plugin_connect_product_to_categories WHERE articleID = 3 AND connect_category_id = $test31ID")->fetchColumn();
```
sbofirov: Please fix all variables with camelCase

Components/CategoryResolver.php
keulinho: exactly
actual less iterations in remove and in add

Components/CategoryResolver.php
keulinho: FUCK!!!!!!!!
Because it made trouble 
the greater than is completly bullshit it has to be less or equals

Tests/Integration/Components/CategoryResolverTest.php
keulinho: yes but the legacy tests don't always use transaction they run before

Tests/Integration/Components/CategoryResolverTest.php
keulinho: it checks that the sql returns the id and verify that the row is there
it would be false if the sql returns no values

Tests/Integration/Components/CategoryResolverTest.php
keulinho: but zero is a valid ID or not?
then you have to cast all IDs returned from executeQuery() to int and that would crash if no result is found becuase then executeQuery returns false

Components/CategoryResolver.php
keulinho: still makes trouble with the correct condition
I created a follow-up:
https://jira.shopware.com/browse/CON-4885

PR #393, opened 2017-09-27
CON-4867: changes are generated when category is deleted

keulinho:


Subscribers/Lifecycle.php
```diff
@@ -167,6 +169,29 @@ public function onDeleteArticle(\Enlight_Event_EventArgs $eventArgs)
...
     /**
+     * Callback function to delete an product from connect
+     * after it is going to be deleted locally
+     *
+     * @param \Enlight_Event_EventArgs $eventArgs
+     */
+    public function onPreDeleteCategory(\Enlight_Event_EventArgs $eventArgs)
+    {
+        $category = $eventArgs->get('entity');
+        $this->connectExport->markProductsInToBeDeletedCategories($category);
```
sbofirov: here you can store category id in private property. ```$this->removedCategoryId = $category->getId()```. then methods are stateful, but cron_update is used for something difference. Then ```$this->connectExport->markProductsInToBeDeletedCategories($category);``` can be removed.

Subscribers/Lifecycle.php
```diff
@@ -167,6 +169,29 @@ public function onDeleteArticle(\Enlight_Event_EventArgs $eventArgs)
...
+
+    /**
+     * Callback function to delete an product from connect
+     * after it is going to be deleted locally
+     *
+     * @param \Enlight_Event_EventArgs $eventArgs
+     */
+    public function onPostDeleteCategory(\Enlight_Event_EventArgs $eventArgs)
+    {
+        $this->connectExport->handleMarkedProducts();
```
sbofirov: In preDelete you already stored categoryId. So in this function you need to check "autoUpdateProducts" config.
```$autoUpdateProducts = $this->config->getConfig('autoUpdateProducts', Config::UPDATE_AUTO);``` can find examples in other subscribers. this config has 3 different values. UPDATE_AUTO means every change of product/variant must be handled. UPDATE_CRON_JOB means changed products will be updated with cron_update flag = 1 and Connect cronjob will generate necessary changes later  and UPDATE_MANUAL means Connect plugin doesn't need to take care of changed products. Customer will update them manually. 

**1.** When values of **autoUpdateProducts** is UPDATE_MANUAL you just need to call return;

**2.** In other 2 cases you need  to fetch all child categories as you did it in onPreDeleteCategory. After that find all related articleIds.

**3.** When values of **autoUpdateProducts** is UPDATE_AUTO then generate changes for all related articleIds ```$this->export($articlesToBeExported);```

**4.** When values of **autoUpdateProducts** is UPDATE_CRON_JOB then update cron_update = 1 for related articledIds.

File: Tests/Integration/Components/ConnectExportTest.php [hunk 3]
```diff
@@ -0,0 +1,141 @@
...
+use Shopware\Components\Model\ModelManager;
+use ShopwarePlugins\Connect\Components\ConfigFactory;
+use ShopwarePlugins\Connect\Components\ConnectExport;
+use ShopwarePlugins\Connect\Components\ErrorHandler;
+use ShopwarePlugins\Connect\Components\Validator\ProductAttributesValidator\ProductsAttributesValidator;
+use ShopwarePlugins\Connect\Tests\DatabaseTestCaseTrait;
+use Shopware\Models\Category\Category;
+use Tests\ShopwarePlugins\Connect\ConnectTestHelper;
+
+class ConnectExportTest extends ConnectTestHelper
```
sbofirov: ConnectTestHelper is not used anymore in new tests structure. So it's not possible integration test to extend it.

Tests/Integration/Components/ConnectExportTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $category = new Category();
+        $category->setId($categoryId);
+
+        $this->connectExport->markProductsInToBeDeletedCategories($category);
+
+        $count = $this->manager->getConnection()->executeQuery('SELECT COUNT(*) FROM `s_plugin_connect_items` WHERE `cron_update` = 1')->fetchColumn();
+        $this->assertEquals(2, (int) $count);
+    }
+
+    public function testMarkProductsInToBeDeletedCategoriesWithChildCategories()
```
sbofirov: why we need 2 different tests for this ? I think testMarkProductsInToBeDeletedCategoriesWithChildCategories also covers testMarkProductsInToBeDeletedCategories. Also you can use fixtures instead of many sql lines in your test.
you can find example in **SwagConnect/Tests/Integration/Components/ProductStream/ProductStreamServiceTest.php** 
```$this->importFixtures(__DIR__ . '/_fixtures/products_and_streams_relation.sql');```

If we decide to handle category deletion in the way that I suggested, some of the tests will be superfluous and new cases must be handled.

Subscribers/Lifecycle.php
keulinho: the lifecycleProvider gets the autoUpdateProducts value from its constructor
no need to call config but i have to take care of that

Tests/Integration/Components/ConnectExportTest.php
keulinho: why not?
it is possible and i use things from it to get the SDK-instance and a helper instance.
Alternative 1: still extend ConnectTestHelper if it's necessary
Alternative 2: create a trait for getting those things
Alternative 3: copy code from ConnectTestHelper into each testcase

Tests/Integration/Components/ConnectExportTest.php
keulinho: good point because fixtures can be reused in lifecycle-subscriber test

Tests/Integration/Components/ConnectExportTest.php
sbofirov: I saw that it's used also in other places as well. If I remember correctly we decided together with Simon that we'll use traits instead of this ConnectTestHelper.

PR #394, opened 2017-09-28
CON-4832: unassign articles from local categories unassignes only current category

keulinho:
If you unassign a product from a local category in Import window,
the article was unassigned from all it's local categories.
This is not the action the user would expect.
Now only the currently selected Category and all child categories got removed.
If the user wants to unassign all categories he can do it over
the edit-function of the Article


Components/ImportService.php
```diff
@@ -544,4 +543,80 @@ private function getAst($categoryId, $query = '', $showOnlyConnectArticles = tru
...
+        $categories = [];
+        $categories[] = $categoryId;
+        $childCategories = $this->manager->getConnection()->executeQuery('SELECT id FROM s_categories WHERE path LIKE ?',
+            ["%|$categoryId|%"]);
+
+        while ($childCategory = $childCategories->fetchColumn()) {
+            $categories[] = (int) $childCategory;
+        }
+
+        $categoriesStatement = $this->manager->getConnection()->prepare('DELETE FROM s_articles_categories WHERE articleID IN (' . implode(', ', $articleIds) . ') AND categoryID IN (' . implode(', ', $categories) . ')');
```
sbofirov: Don't know why, but doesn't work when try to run it manually.
I saw that tests are reaching this line.
have you ran it in your shopware?
I'm trying to run ```EXPLAIN DELETE FROM s_articles_categories WHERE articleID IN (5,1,2,3) AND categoryID IN (5,1,2,3);``` and got following error: 10:25:51 Kernel error: Error( 1064 ) 42000: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DELETE FROM s_articles_categories WHERE articleID IN (5,1,2,3) AND categoryID IN' at line 1".

if you want to use where IN, you need to do nested queries I think.
Btw 
```
if ($categoryId !== null) {
 +                    $this->unAssignArticlesFromCategory($articleIds, $categoryId);
 +                } else {
 +                    $this->unAssignArticlesFromAllCategories($articleIds);
 +                }
``` 
is wrapped in transaction. You know articleIds and categoryIds including child categories. Then you can call delete query for combination of every article & category. At the end they'll be commited at once.

Components/ImportService.php
```diff
@@ -544,4 +543,80 @@ private function getAst($categoryId, $query = '', $showOnlyConnectArticles = tru
...
+        $this->manager->getConnection()->executeQuery('
+                  DELETE FROM s_articles_categories_ro
+                  WHERE parentCategoryID NOT IN 
+                    (SELECT categoryID
+                      FROM s_articles_categories
+                      WHERE s_articles_categories.articleID = s_articles_categories_ro.articleID
+                    )
+                  ');
+
+
```
sbofirov: not needed empty line. looks like cs-fixer is not working on your machine. Did you install githooks ?

Components/ImportService.php
```diff
@@ -544,4 +543,80 @@ private function getAst($categoryId, $query = '', $showOnlyConnectArticles = tru
...
+                  ');
+
+        //we have to check the parent category too, because we can have following entries in s_articles_category_id:
+        //|articleID|categoryID|parentCategoryID
+        //|        1|         3|              5
+        //|        1|         3|              6
+        // -> category 6 got deleted from article but isn't delete in previous statement because categoryID is valid
+        $this->manager->getConnection()->executeQuery('
+                  DELETE FROM s_articles_categories_ro
+                  WHERE parentCategoryID NOT IN
```
sbofirov: Same here. I think NOT IN is not good decision for table with so many rows.

I guess you meant s_articles_categories_ro. How is possible to   have 2 different parent categories for same category ? If we use ```DELETE FROM s_articles_categories_ro WHERE articleID = ? AND categoryID = ?``` then this query is superfluous.

Components/ImportService.php
```diff
@@ -544,4 +543,80 @@ private function getAst($categoryId, $query = '', $showOnlyConnectArticles = tru
...
+
+        $categoriesStatement = $this->manager->getConnection()->prepare('DELETE FROM s_articles_categories WHERE articleID IN (' . implode(', ', $articleIds) . ') AND categoryID IN (' . implode(', ', $categories) . ')');
+        $categoriesStatement->execute();
+
+        //Deletes all entries in s_articles_categories_ro that are not in the Path to a assigned category
+        //need the nested select because mysql got an error otherwise
+        //see: https://stackoverflow.com/questions/45494/mysql-error-1093-cant-specify-target-table-for-update-in-from-clause
+        $this->manager->getConnection()->executeQuery('
+                  DELETE FROM s_articles_categories_ro
+                  WHERE id NOT IN
```
sbofirov: I think NOT IN is too dangerous. In my shop I've 140K products. Most of them don't have any category, but usually every article has more than 1 category. So s_articles_categories_ro will contain more than 140K rows. Then 2 nested selects with 2 inner joins and LIKE is good enough. Actually why we need this query? As I said your already know articleIds and categoryIds.  then we can remove them one by one.

Components/ImportService.php
```diff
@@ -544,4 +543,80 @@ private function getAst($categoryId, $query = '', $showOnlyConnectArticles = tru
...
+                  WHERE parentCategoryID NOT IN 
+                    (SELECT categoryID
+                      FROM s_articles_categories
+                      WHERE s_articles_categories.articleID = s_articles_categories_ro.articleID
+                    )
+                  ');
+
+
+        $attributeStatement = $this->manager->getConnection()->prepare('
+                    UPDATE s_articles_attributes
```
sbofirov: this is tricky, because we need to check which articleIds are not assigned to any remote category, but I'd suggest to use something else instead of NOT IN. Probably you can use COUNT(*) with GROUP BY as we did it yesterday. add where in and already known articleIds. then you'll receive list of articleIds and how many categories they have assigned. Keep in mind that you need to count only remote categories.

Models/Connect/ProductToRemoteCategoryRepository.php
```diff
@@ -97,6 +97,9 @@ public function findArticleIdsByRemoteCategory($remoteCategoryKey)
         $builder->leftJoin('a.attribute', 'att');
         $builder->where('rc.categoryKey = :categoryKey');
         $builder->setParameter('categoryKey', $remoteCategoryKey);
+        //distinct necessary because of variant articles
+        //each variant has an own entry in a.attribute so same articleId iis returned multiple times
+        $builder->distinct();
```
sbofirov: :+1:

Components/ImportService.php
keulinho: I can run it locally 
image.
I don't understand what you mean in the second part

Components/ImportService.php
keulinho: the column parent in s_articles_categories_ro is actually the child category -> i had a discussion with jan bücker from the core about this one
to extend the example above
s_articles_categories_ro contains 
|articleID|categoryID|parent|
|           1|               1|      1.1|
|           1|               1|      1.2|

if 1.2 gets unassigned then second row has to be deleted but first one has to stay

Components/ImportService.php
keulinho: The problem is in s_articles_categories_ro are all parent categories assigned to the articles stored
e.g 
i have categories tree like:
1:
    1.1:
        1.1.1
        1.1.2
    1.2
In s_articles_categories i have:
|articleID|categoryID|
           1 | 1.1.1
           1 | 1.2
this leads to s_articles_categories_ro also containing categories 1 and 1.1 for article 1

Components/ImportService.php
keulinho: the githooks are running
but i'll fix that

PR #398, opened 2017-10-17
CON-4922 Refactor storefront usage in stream export

sbofirov:
Storefront namespace is designed to be used only in shopware frontend,
but we need to fetch articles/detal by stream id during stream export in backend.
Problem is that ShopContext is using Shop entity and it will break current session in DB.
With this PR Storefront namespace is not used anymore as a dependency.

### DIFF
--- a/Subscribers/CronJob.php
+++ b/Subscribers/CronJob.php
@@ -65,23 +65,32 @@ class CronJob implements SubscriberInterface
     private $container;
 
     /**
+     * @var ProductStreamService
+     */
+    private $productStreamService;
+
+    /**
      * @param SDK $sdk
      * @param ConnectExport $connectExport
      * @param Config $configComponent
+     * @param Helper $helper
      * @param Container $container
+     * @param ProductStreamService $productStreamService
      */
     public function __construct(
         SDK $sdk,
         ConnectExport $connectExport,
         Config $configComponent,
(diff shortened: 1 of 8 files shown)


Subscribers/CronJob.php
```diff
@@ -228,9 +243,26 @@ private function removeArticlesFromStream(array $articleIds)
...
+     */
+    private function getProductSearch()
+    {
+        if (!$this->productSearch) {
+            // HACK
+            // do not use as a dependency!!!
+            // this class uses Shopware product search which depends on shop context
+            // so if it's used as dependency of subscriber, plugin returns error on deactivate
+            // see CON-4922
+            $this->productSearch = $this->container->get('swagconnect.product_search');
```
keulinho: if this works fine -> but shoudn't the streamservice and thumbnail manager be injected in the constructor?
Imho it's really bad to use the container because your dependencies are implicit so i would suggest to put as much in the constructor as possible

sbofirov:
@keulinho absolutely agree with you. I changed ProductStreamService to be used as a dependency, but for thumbnail_manager it's not possible, because it also uses Shop entity in constructor and this can break deactivation as well.

We need to move to new plugin system asap, because it will help us to use DI container in better way, with services.xml

PR #399, opened 2017-10-18
[Con 4873] Enable "Export All" button to export lots of products

Christian-Rades:
Implemented the "export all action" as a batch process to handle big amounts without gui tmeouts


Views/backend/connect/view/export/product/many_products_dialog.js
```diff
@@ -25,6 +25,11 @@ Ext.define('Shopware.apps.Connect.view.export.product.manyProductsDialog', {
     sourceIds: [],
 
     /**
+     * Contains the amount of products export all will export as batches
+     */
+    count: 0,
```
keulinho: why a global variable == global state? can you refactor it so that this is passed as argument

Views/backend/connect/view/export/product/progress.js
```diff
@@ -30,6 +30,16 @@ Ext.define('Shopware.apps.Connect.view.export.product.Progress', {
...
 
     /**
+     * if this is true this will export all products as batches
+     */
+    exportAll: false,
+
+    /**
+     * Contains the amount of products export all will export as batches
+     */
+    count: 0,
```
keulinho: same here and above exportAll too

Views/backend/connect/controller/main.js
```diff
@@ -1096,20 +1157,21 @@ Ext.define('Shopware.apps.Connect.controller.Main', {
...
-
                 if (!operation.success) {
                     me.createGrowlMessage(title, operation.message, true);
                 } else {
-                    if (operation.sourceIds.length > 1000) {
+                    if (operation.count > 1000) {
 
                         Ext.create('Shopware.apps.Connect.view.export.product.manyProductsDialog', {
-                            sourceIds: operation.sourceIds
+                            exportAll: true,
```
keulinho: Imho these bool switches are also bad practice
maybe we can find a cleaner solution here

Views/backend/connect/view/export/product/progress.js
```diff
@@ -89,12 +100,15 @@ Ext.define('Shopware.apps.Connect.view.export.product.Progress', {
                 }
                 me.startButton.setDisabled(true);
                 me.cancelButton.setDisabled(true);
-
-                me.fireEvent('startExport', me.sourceIds, me.batchSize, me);
+                if (me.exportAll) {
```
keulinho: i think the progress bar shouldn't fire any event at all
it should just show some progress that will be updated but shouldn't know anything more
i would suggest firing this event after creating this progess bar

PR #403, opened 2017-10-27
CON-4938: added Exception catch if media entity was deleted

keulinho:
theres no cascade delete for media and image entity.
problem was that an exception was thrown
replaced an if try ... catch


keulinho:
Somehow the failing test somehow depends on the ImageImporterTest 
befor the ImageImporterTest was running before the LegacyLifecycleTest
after i moved it to Integration it runs after the LegacyLifecycleTest
At the moment the LifecycleTest receives just one change
That one is an insert change and the price in the payload is 123.99 
that is the price set in LifecycleTest::insertVariants()

Tests/Legacy/Shopware/Connect/Subscribers/LifecycleTestLegacy.php
```diff
@@ -12,10 +12,10 @@
 use ShopwarePlugins\Connect\Components\ConnectExport;
 use ShopwarePlugins\Connect\Components\Validator\ProductAttributesValidator\ProductsAttributesValidator;
 use Tests\ShopwarePlugins\Connect\ConnectTestHelper;
-use ShopwarePlugins\Connect\Subscribers\Lifecycle;
 use ShopwarePlugins\Connect\Components\ErrorHandler;
 
-class LifecycleTest extends ConnectTestHelper
+/** Please don't rename it because this test will fail if the classname is LifecycleTest */
+class LifecycleTestLegacy extends ConnectTestHelper
```
sbofirov: Then the test is not executed at all. I tried to put var_dump('something');exit; in testUpdatePrices and it doesn't work if you run ./psh.phar local:unit command

Tests/Integration/Components/ImageImportTest.php
```diff
@@ -0,0 +1,193 @@
...
+    private $manager;
+    /** @var ImageImport */
+    private $imageImport;
+
+    public function SetUp()
+    {
+        $this->manager = Shopware()->Models();
+        $this->imageImport = new ImageImport(
+            Shopware()->Models(),
+            Shopware()->Plugins()->Backend()->SwagConnect()->getHelper(),
```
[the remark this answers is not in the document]
keulinho: yes but after using the productBuilderTrait this would be the only usage of the connectTestHelperTrait 
i decided to use this and not depend on the ConnectTestHelperTrait at all

Tests/Functional/Subscribers/LifecycleTest.php
```diff
@@ -15,9 +17,21 @@ class LifecycleTest extends \Enlight_Components_Test_Plugin_TestCase
     use DatabaseTestCaseTrait;
 
     /**
+     * @param string $file
+     */
+    public function importFixtures($file)
```
sbofirov: ```DatabaseTestCaseTrait``` already has method ```importFixtures``` you can use it

PR #405, opened 2017-10-31
CON-4926 Added index by articleID in product_to_categories

sbofirov:
takes 1.6 sec if articleID exists or even 5 seconds when execute following query in customer db:
```
SELECT *
FROM `s_plugin_connect_product_to_categories`
WHERE articleID = 999999;
```


Models/Connect/ProductToRemoteCategoryRepository.php
```diff
@@ -135,7 +135,7 @@ public function getRemoteCategoryIds($articleId)
         $builder = $this->createQueryBuilder('ptrc');
         $builder->select('ptrc.connectCategoryId');
         $builder->where('ptrc.articleId = :articleId');
-        $builder->setParameter('articleId', $articleId);
+        $builder->setParameter('articleId', (int) $articleId);
```
keulinho: is this enough to make doctrine do an int compare or do you need?:
`$builder->setParameter('articleId', (int) $articleId, PDO::PARAM_INT);`

Bootstrap/Update.php
```diff
@@ -483,4 +484,23 @@ private function recreateRemoteCategoriesAndProductAssignments()
             }
         }
     }
+
+    /**
+     * Create index by articleID in s_plugin_connect_product_to_categories table.
+     */
+    private function addProductToCategoryIndex()
+    {
+        if (version_compare($this->version, '1.1.6', '<=')) {
```
keulinho: we released 1.1.7 today so it has to be 1.1.7

plugin.json
```diff
@@ -7,7 +7,7 @@
     "license": "proprietary",
     "link": "http://store.shopware.com",
     "author": "Shopware AG",
-    "currentVersion": "1.1.6",
+    "currentVersion": "1.1.7",
```
keulinho: you have to increase it to 1.1.8

Models/Connect/ProductToRemoteCategoryRepository.php
```diff
@@ -158,8 +158,8 @@ public function deleteByConnectCategoryId($categoryId, $articleId)
         $builder->delete('Shopware\CustomModels\Connect\ProductToRemoteCategory', 'ptrc');
         $builder->where('ptrc.connectCategoryId = :ccid');
         $builder->andWhere('ptrc.articleId = :articleId');
-        $builder->setParameter(':ccid', $categoryId);
-        $builder->setParameter(':articleId', $articleId);
+        $builder->setParameter(':ccid', (int) $categoryId);
+        $builder->setParameter(':articleId', (int) $articleId);
```
keulinho: \PDO::PARAM_INT here and the line above too
Sorry i missed that in the first review

PR #410, opened 2017-11-09
CON-4885: empty categories gets deleted

keulinho:
if an article is no longer part of an connect
category the article will be deleted from
all local categories that are related to that
connect category
if the connect category is empty afterwards it
is deleted
if an article gets deleted through a change
it gets checked if any of the categories
of the article have to be deleted


Components/CategoryResolver.php
```diff
@@ -129,12 +137,81 @@ private function removeProductsFromNotAssignedRemoteCategories(array $assignedCa
...
+    private function deleteAssignmentOfLocalCategories($currentProductCategoryId, $articleId)
+    {
+        $localCategoriesIds = $this->manager->getConnection()->executeQuery(
+            'SELECT local_category_id FROM s_plugin_connect_categories_to_local_categories WHERE remote_category_id = ?',
+            [$currentProductCategoryId]
+        )->fetchAll(\PDO::FETCH_COLUMN);
+        if ($localCategoriesIds) {
+            foreach ($localCategoriesIds as $categoryId) {
+                $this->manager->getConnection()->executeQuery(
+                    'DELETE FROM `s_articles_categories` WHERE `articleID` = ? AND `categoryID` = ?',
```
sbofirov: probably you can do it with where in. I don't expect too many items in ```$localCategoriesIds```

Components/CategoryResolver.php
```diff
@@ -129,12 +137,81 @@ private function removeProductsFromNotAssignedRemoteCategories(array $assignedCa
...
+     * @param int $categoryId
+     */
+    private function deleteEmptyCategory($categoryId)
+    {
+        $connectImported = $this->manager->getConnection()->fetchColumn(
+            'SELECT connect_imported_category FROM s_categories_attributes WHERE categoryID = ?',
+            [$categoryId]
+        );
+
+        $childCount = (int) $this->manager->getConnection()->fetchColumn(
```
sbofirov: you can extract this query in private method. then in if statement below 
```if ($connectImported == 1 && $this->countChildCategories($categoryId) === 0) {```

By this way the query won't be executed when ```$categoryId``` is not imported from connect

File: Tests/Unit/Subscribers/ServiceContainerTest.php [hunk 3]
```diff
@@ -106,6 +107,10 @@ public function testOnAutoCategoryResolver()
                 [ProductToRemoteCategory::class, $this->createMock(ProductToRemoteCategoryRepository::class)],
             ]));
 
+        $this->diContainer
+            ->method('get')
```
sbofirov: better add ```->with('CategoryDenormalization')```. right now will return CategoryDenormalization always, doesn't matter what's the argument.

Tests/Unit/Subscribers/ServiceContainerTest.php
```diff
@@ -119,6 +124,10 @@ public function testOnDefaultCategoryResolver()
...
```
sbofirov: same here

PR #411, opened 2017-11-10
CON-4955: ArticleSubscriber uses right FetchMode

keulinho:
before the ArticleSubscriber used ->fetchColumn
that returns the result of the first row as string
after that we called export() with that
result
but exports() expects an array as input
with executeQuery()->fetchAll(\PDO::FETCH_COLUMN)
this query will return an array of articleIds

### DIFF
--- a/Subscribers/Article.php
+++ b/Subscribers/Article.php
@@ -230,10 +230,10 @@ class Article implements SubscriberInterface
                     $this->modelManager->flush();
                 }
 
-                $sourceIds = $this->modelManager->getConnection()->fetchColumn(
+                $sourceIds = $this->modelManager->getConnection()->executeQuery(
                     'SELECT source_id FROM s_plugin_connect_items WHERE article_id = ?',
                     [$article->getId()]
-                );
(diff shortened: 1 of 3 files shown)


Subscribers/Article.php
```diff
@@ -230,10 +230,10 @@ public function extendBackendArticle(\Enlight_Event_EventArgs $args)
                     $this->modelManager->flush();
                 }
 
-                $sourceIds = $this->modelManager->getConnection()->fetchColumn(
+                $sourceIds = $this->modelManager->getConnection()->executeQuery(
                     'SELECT source_id FROM s_plugin_connect_items WHERE article_id = ?',
                     [$article->getId()]
-                );
+                )->fetchAll(\PDO::FETCH_COLUMN);
```
sbofirov: :+1: 
This will fetch all sourceIds by given articleId and will work for variants.
I'm wondering only if the article has 4-5K variants. Maybe we need to check if the articles must be updated via CRON just update column cron_update.
@keulinho what do you think.

PR #412, opened 2017-11-13
CON-4954: added shopId to ConnectCategories

keulinho:
before Connect Categories don't had the shopId
that made problems when two suppliers had the
same category with the same path
to distingush this we introduced a new field
shop_id in s_plugin_connect_categories
ef7b0ad


Components/ImportService.php
```diff
@@ -332,16 +332,18 @@ public function storeRemoteCategories(array $remoteItems)
 
         $connection->beginTransaction();
         try {
-            foreach ($remoteItems as $articleId => $categories) {
+            foreach ($remoteItems as $articleId => $remoteItem) {
+                $categories = $remoteItem[0];
+                $shopId = $remoteItem[1];
```
sbofirov: maybe it's time to create an object with following properties:
* articleId
* categories array
* shopId

Actually you can use SDK Product. ```$remoteItems[$articleId] = new Product```.

It's hard to work with arrays and indexes. Always need to check if the index exists.

Components/Helper.php
```diff
@@ -878,4 +894,86 @@ public function getProductCountForCategoryRecovery()
...
+
+        while ($row = $result->fetch()) {
+            $categories = json_decode($row['category'], true);
+            $this->manager->getConnection()->executeQuery('DELETE FROM s_plugin_connect_product_to_categories WHERE articleID = ?',
+                [$row['article_id']]
+            );
+
+            $shopId = (int) $row['shop_id'];
+            foreach ($categories as $categoryKey => $category) {
+                $selectedCategory = $this->manager->getConnection()->executeQuery('SELECT `id` FROM s_plugin_connect_categories WHERE category_key = ? AND shop_id = ?',
```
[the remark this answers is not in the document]
keulinho: no i can't simply do that in one query because first i check if that connect category with that specific shopId exists
if this is the case i don't need to update ConnectCategories
in the second query with shopId = null i check for the "old" categories without shopId then i can simply update that row and add the shopId so i don't need to insert a new row

but nevertheless i extracted the code to make it more readable

Components/Helper.php
```diff
@@ -866,4 +882,86 @@ public function getProductCountForCategoryRecovery()
...
+        return (int) $res['id'];
+    }
+
+    /**
+     * @param int $createdCategoryId
+     * @param string $categoryKey
+     */
+    private function assignLocalCategories($createdCategoryId, $categoryKey)
+    {
+        $originalCategoryId = $this->manager->getConnection()->fetchColumn('SELECT `id` FROM s_plugin_connect_categories WHERE category_key = ? AND id <> ?',
```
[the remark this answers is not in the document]
keulinho: in fact I do only insert new rows if it is necessary
in the comment above i explained the thing for s_connect_categories

for s_plugin_connect_categories_to_local_categories:
if i have found an s_connect_category without shop_id this method isn't executed tho i don't change s_plugin_connect_categories_to_local_categories
if i have to create a new category entry this function fetches all local categories that are assigned to the "original" connect_category and assign them to the new created category to -> so we don't break things
in fact if we have to create a new entry in s_plugin_connect_categories and the old category was assigned to a local category we don't know from wich supplier this should be so we simply assign all categories with the same path

Components/ImportService.php
keulinho: this functions is completely duplicated by CategoryResolver::storeRemoteCategories
in categoryResolver it has proper input parameter

Components/ImportService.php
```diff
@@ -399,8 +389,11 @@ public function getArticlesWithAutoImportedCategories()
         foreach ($statement->fetchAll(\PDO::FETCH_ASSOC) as $item) {
             $categories = json_decode($item['category'], true);
             if (is_array($categories) && count($categories) > 0) {
-                $articleId = $item['article_id'];
-                $remoteItems[$articleId] = [$categories, $item['shop_id']];
+                $product = new Product();
+                $product->shopId = $item['shop_id'];
+                $product->sourceId = $item['article_id'];
+                $product->categories = $categories;
+                $remoteItems[] = $product;
```
sbofirov: I prefer to use ```$remoteItems[$articleId] = $product```. Then ```$product->sourceId = $item['article_id'];``` is not needed.

sourceId confusing me because it contains articleId.

Components/Helper.php
```diff
@@ -924,6 +907,26 @@ public function addShopIdToConnectCategories($offset, $batchsize)
...
+        $selectedCategory = $this->manager->getConnection()->executeQuery('SELECT `id` FROM s_plugin_connect_categories WHERE category_key = ? AND shop_id = ?',
+            [$categoryKey, $shopId]);
+        if (!($res = $selectedCategory->fetch())) {
+            $categoryId = $this->createCategoryWithShopId($categoryKey, $shopId, $category);
+        } else {
+            $categoryId = (int)$res['id'];
+        }
+        $this->manager->getConnection()->executeQuery('INSERT INTO s_plugin_connect_product_to_categories (articleID, connect_category_id) VALUES (?, ?)',
+            [$row['article_id'], $categoryId]
+        );
```
sbofirov: I meant to move all the logic from this private method to ```createCategoryWithShopId```
Then you can use ``` $selectedCategory = $this->manager->getConnection()->executeQuery('SELECT `id` FROM s_plugin_connect_categories WHERE category_key = ? AND shop_id IS NULL',
            [$categoryKey]);```  only once. Don't need to fetch it by category key and shop id. If it doesn't exist then call ```$this->createCategoryWithShopId($categoryKey, $shopId, $category);```.

My idea is to call the query above then if category doesn't exist or it has different shopID insert new record in s_plugin_connect_categories

keulinho: But i have to fetch this by shopId and categoryKey to know if it already exists
it's important that i fetch one time with the exact shopId and one time sourceId should be null
that are two different cases i can't do it in one query

Components/ImportService.php
keulinho: yes good point that confused me too
i wasn't happy with it either

PR #413, opened 2017-11-14
 CON-4896: check if cronPlugin is activated in import settings 

keulinho:
when the user tries to load the images with cronjob it is checked that the
Cronjob Plugin is activated if not an error message is shown
and the config value can't be changed
if cron Plugin is activated the SwagConnectImageCronJob is
automatically activated when the settig is set to import images with Cron
accordingly the cronjob gets automatically deactivated if images
are imported on first import.


Tests/Integration/Components/ConfigTest.php
```diff
@@ -5,53 +5,15 @@
...
-namespace Tests\ShopwarePlugins\Connect;
+namespace ShopwarePlugins\Connect\Tests\Integration\Components;
 
 use ShopwarePlugins\Connect\Components\ConfigFactory;
 use ShopwarePlugins\Connect\Components\Marketplace\MarketplaceSettings;
-use ShopwarePlugins\Connect\Components\ProductQuery\BaseProductQuery;
+use ShopwarePlugins\Connect\Tests\DatabaseTestCaseTrait;
 
-class ConfigTest extends ConnectTestHelper
+class ConfigTest extends \PHPUnit_Framework_TestCase
```
sbofirov: :+1:

Views/backend/connect/view/config/import/form.js
```diff
@@ -219,12 +225,34 @@ Ext.define('Shopware.apps.Connect.view.config.import.Form', {
...
                             listeners:{
                                 change: function(checkbox, newValue, oldValue, opts){
+                                    if (checkbox.getValue() === false) {
+                                        Ext.Ajax.request({
+                                            url: '{url controller=ConnectConfig action=checkCronPlugin}',
+                                            method: 'GET',
+                                            success: function (response, opts) {
+                                                var data = Ext.JSON.decode(response.responseText);
+                                                if (data.cronActivated !== true) {
+                                                    checkbox.setValue(true);
```
sbofirov: so if the cron plugin is not active I'm not able to uncheck option "Import images during initial import?"

Components/Config.php
```diff
@@ -362,6 +362,12 @@ public function setImportConfigs($data)
                     $model->setValue($configValue);
                 }
                 $this->manager->persist($model);
+
+                if ($key === 'importImagesOnFirstImport' && $configValue === 0) {
```
sbofirov: what about export products with cronjob when cron plugin is not active? This case is described in jira.

keulinho: oh fuck i somehow missed that completely 
I will add that functionality for update Products too

Views/backend/connect/view/config/import/form.js
keulinho: exactly because if deactivating this option without having CronPlugin installed it would be like saying "I don't want to import images" and that is probably not what the user wants
if someone wants that i think we need to make an explicit option for that

PR #414, opened 2017-11-16
CON-4928: Added a Functional test for the main Connect Controller 

Christian-Rades:
Added the WebTestCaseTrait from the B2B suite and its infrastructure.
Implemented a ConnectControllerTest to check the most used actions.
Refactored all functional tests to use the WebTestCaseTrait


Tests/AbstractConnectUnitTest.php
```diff
@@ -33,7 +33,7 @@ protected function tearDown()
     public function killShopwareFunctionBefore()
     {
         self::$appState = Shopware();
-        Shopware(new EmptyShopwareApplication());
+        Shopware(new EmptyShopwareApplication2());
```
keulinho: why the 2 after the classname?

Tests/AbstractConnectUnitTest.php
```diff
@@ -45,7 +45,7 @@ public function restoreShopwareFunctionAfter()
     }
 }
 
-class EmptyShopwareApplication
+class EmptyShopwareApplication2
```
keulinho: same here

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+    use WebTestCaseTrait;
+
+    /**
+     * @param Response $response
+     * @return array
+     */
+    private function handleJsonResponse(Response $response)
+    {
+        $this->assertEquals(200, $response->getStatusCode());
+        $responseData = \json_decode($response->getContent(), true);
```
keulinho: whats the backslash for?

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Functional\Controller;
+
+use ShopwarePlugins\Connect\Tests\WebTestCaseTrait;
+use ShopwarePlugins\Connect\Tests\TestClient;
+use Symfony\Component\HttpFoundation\Response;
+
+class ConnectTest extends \PHPUnit_Framework_TestCase
```
keulinho: in generally this whole class looks very good and it is great to have such testCases
but wouldn't it be better to check not just the json response but also the database for request that change the db?
so we would have the backend tested end-to-end from api to the db

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+        /** @var TestClient $client */
+        $client = self::createBackendClient();
+
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/config_fixes.sql');
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/connect_items.sql');
+
+        $client->request('POST', 'backend/Connect/insertOrUpdateProduct', ['sourceIds' => ['2','2-123','2-124']]);
+        $responseData = $this->handleJsonResponse($client->getResponse());
+
+        $this->assertTrue($responseData['success']);
```
keulinho: here it would make sense to check wether the export_status in s_plugin_connect_items of the products with the given source ids is 'update'

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    public function test_error_handling_on_export()
+    {
+        /** @var TestClient $client */
+        $client = self::createBackendClient();
+
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/connect_items.sql');
+
+        $client->request('POST', 'backend/Connect/insertOrUpdateProduct', ['sourceIds' => ['2','2-123']]);
```
keulinho: maybe test here one product without error, that not all exports will fail if one pricefield is not ok

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $responseData = $this->handleJsonResponse($client->getResponse());
+
+        $this->assertFalse($responseData['success']);
+        $this->assertCount(2, $responseData['messages']['price']);
```
keulinho: here ypou can check db for error messages to, and if you test is one "good" product assert the same as in the testcase above in the db

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/config_fixes.sql');
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/connect_items.sql');
+        $this->importFixturesFileOnce(__DIR__ . '/_fixtures/sw_product_stream.sql');
+
+        $streamId = self::getDbalConnection()->fetchColumn('SELECT id FROM s_product_streams WHERE name = ?', ['AwStream']);
+
+        $client->request('POST', 'backend/Connect/exportStream', ['offset' => 0, 'limit' => 50, 'streamIds' => [$streamId]]);
+        $responseData = $this->handleJsonResponse($client->getResponse());
+
+        $this->assertTrue($responseData['success']);
```
keulinho: same here please check db

Tests/Functional/Controller/ImportTest.php
```diff
@@ -7,185 +7,232 @@
...
-use ShopwarePlugins\Connect\Tests\DatabaseTestCaseTrait;
+use ShopwarePlugins\Connect\Tests\TestClient;
+use ShopwarePlugins\Connect\Tests\WebTestCaseTrait;
+use Symfony\Component\HttpFoundation\Response;
 
-class ImportTest extends \Enlight_Components_Test_Plugin_TestCase
+class ImportTest extends \PHPUnit_Framework_TestCase
 {
-    use DatabaseTestCaseTrait;
+    use WebTestCaseTrait;
```
keulinho: does the webTestCaseTrait roll back changes to the db ->
because if you invoke the backend controllers something could be changed in the db 
even if you do no db-queries in the testCase

Tests/Functional/Controller/ImportTest.php
```diff
@@ -7,185 +7,232 @@
...
 
-    public function setUp()
+    /**
+     * @param Response $response
+     * @return array
+     */
+    private function handleJsonResponse(Response $response)
```
keulinho: this could be part of the WebTestCase trait or not?
why copy it all over every testclass

Tests/Functional/Controller/ImportTest.php
```diff
@@ -7,185 +7,232 @@
...
+            [
+                'localCategoryId' => 140809703,
+                'remoteCategoryKey' => '/deutsch/television',
+                'remoteCategoryLabel' => 'Television',
+                'node' => 'shopId6~stream~Awesome Products~/deutsch/television'
+            ]
+        );
+        $responseData = $this->handleJsonResponse($client->getResponse());
+
+        $this->assertTrue($responseData['success']);
```
keulinho: here again a db check would make sense

Tests/Functional/Controller/ImportTest.php
Christian-Rades: Yes the KernelTestCase already wraps everything into a transaction and the WebTestCase is just an extension

Tests/Functional/Controller/ImportTest.php
```diff
@@ -7,185 +7,232 @@
...
-        $this->dispatch('backend/Import/unassignRemoteArticlesFromLocalCategory');
+        /** @var TestClient $client */
+        $client = $this->createBackendClient();
 
-        self::assertEquals(200, $this->Response()->getHttpResponseCode());
-        self::assertTrue($this->View()->success);
+        $client->request('POST', 'backend/Import/unassignRemoteArticlesFromLocalCategory');
+        $responseData = $this->handleJsonResponse($client->getResponse());
+
+        $this->assertTrue($responseData['success']);
```
keulinho: please check db

Tests/Functional/Controller/ImportTest.php
```diff
@@ -7,185 +7,232 @@
...
-        $this->dispatch('backend/Import/recreateRemoteCategories');
+        /** @var TestClient $client */
+        $client = $this->createBackendClient();
+
+        $client->request('POST', 'backend/Import/recreateRemoteCategories');
+        $responseData = $this->handleJsonResponse($client->getResponse());
 
-        self::assertEquals(200, $this->Response()->getHttpResponseCode());
-        self::assertTrue($this->View()->success);
+        $this->assertTrue($responseData['success']);
```
keulinho: same here

Tests/Functional/Subscribers/LifecycleTest.php
```diff
@@ -7,44 +7,44 @@
...
+use Shopware\Models\Article\Article;
+use Shopware\Models\Article\Detail;
+use ShopwarePlugins\Connect\Tests\TestClient;
+use ShopwarePlugins\Connect\Tests\WebTestCaseTrait;
+use Symfony\Component\HttpFoundation\Response;
+
+class LifecycleTest extends \PHPUnit_Framework_TestCase
 {
-    use DatabaseTestCaseTrait;
+    use WebTestCaseTrait;
```
keulinho: does the webTestCaseTrait roll back changes to the db ->
because if you invoke the backend controllers something could be changed in the db
even if you do no db-queries in the testCase

Tests/Functional/Subscribers/LifecycleTest.php
```diff
@@ -7,44 +7,44 @@
...
+     * @param Response $response
+     * @return array
      */
-    private $manager;
-
-    /**
-     * @before
-     */
-    public function prepare()
+    private function handleJsonResponse(Response $response)
```
keulinho: move it to trait

Tests/Functional/Subscribers/LifecycleTest.php
```diff
@@ -7,44 +7,44 @@
...
+        $client->request('POST', 'backend/Article/deleteDetail', ['articleId' => 32870, 'id' => 2404537]);
 
-        $this->assertEquals(200, $this->Response()->getHttpResponseCode());
-        $this->assertTrue($this->View()->success);
+        $responseData = $this->handleJsonResponse($client->getResponse());
 
-        $changes = $this->manager->getConnection()->fetchAll(
+        $this->assertTrue($responseData['success']);
+
+        $changes = self::getDbalConnection()->fetchAll(
```
keulinho: i mean db checks just like this that verify that it exactly works
just because the controller returns success it doesn't mean that the controller worked as expected

Tests/KernelTestCaseTrait.php
```diff
@@ -0,0 +1,313 @@
...
+
+    /**
+     * @var bool
+     */
+    private $disableCommonFixtures = false;
+
+    /**
+     * @var array
+     */
+//    private $commonTestFixtures = [
```
keulinho: can be removed if it is outcommented

Tests/KernelTestCaseTrait.php
```diff
@@ -0,0 +1,313 @@
...
+        };
+
+        $auth = \Shopware_Components_Auth::getInstance();
+        $auth->setBaseAdapter($adapter);
+        $auth->addAdapter($adapter);
+
+        self::getKernel()->getContainer()->set('auth', $auth);
+    }
+
+    public function performB2bContactLogin()
```
keulinho: i think we don't need these b2b specific methods 
please remove evrything that we don't need

Tests/Functional/Controller/ImportTest.php
Christian-Rades: I am not sure putting about putting assert calls into a trait, because that implies that the trait resides in a subclass of PHPUnit_Framework_TestCase

Tests/TestKernel.php
```diff
@@ -0,0 +1,168 @@
...
+        {
+            \Smarty_Resource::$sources = [];
+            \Smarty_Resource::$compileds = [];
+
+            $this->container->reset('front');
+            \Enlight_Template_Manager::$_smarty_vars = [];
+        }
+    }
+
+    class EmptyShopwareApplication
```
keulinho: why are here multiple classes in one file?
please split it to different files if it's possible

Tests/TestKernel.php
```diff
@@ -0,0 +1,168 @@
...
+         */
+        private function throwException($type, $name)
+        {
+            throw new \DomainException('Restricted to ' . $type . ' ' . $name . ' on Shopware() , because you should not have a kernel in this test case.');
+        }
+    }
+}
+
+namespace {
+    class Measurement
```
keulinho: same as above

Tests/TestKernel.php
Christian-Rades: This class is simply a stub to keep the following tests rom using the kernel ater the functional tests ran. It's only used this once and wont be modified in the future, so i think it would just pollute the file system

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -70,24 +56,40 @@ public function test_exporting_a_product_with_variants()
...
-        $client->request('POST', 'backend/Connect/insertOrUpdateProduct', ['sourceIds' => ['2','2-123','2-124']]);
+        $sourceIds = ['2','2-123','2-124'];
+
+        $client->request('POST', 'backend/Connect/insertOrUpdateProduct', ['sourceIds' => $sourceIds]);
         $responseData = $this->handleJsonResponse($client->getResponse());
 
         $this->assertTrue($responseData['success']);
+
+        $exportedItems = self::getDbalConnection()->fetchAll(
+            "SELECT export_status FROM s_plugin_connect_items WHERE source_id in (?, ?, ?) AND export_status = 'update' ",
```
keulinho: generally this works
but you can use
'SQL with WHERE x IN (?)',
[sourceIds],
\Doctrine\DBAL\Connection::PARAM_INT_ARRAY

this will work for arrays with any length even if you don't know the length beforehand
Also please capitalize IN it is an mysql keyword

Tests/Functional/Controller/ConnectTest.php
```diff
@@ -70,24 +56,40 @@ public function test_exporting_a_product_with_variants()
...
+
         $client->request('POST', 'backend/Connect/insertOrUpdateProduct', ['sourceIds' => ['2','2-123']]);
         $responseData = $this->handleJsonResponse($client->getResponse());
 
         $this->assertFalse($responseData['success']);
-        $this->assertCount(2, $responseData['messages']['price']);
+        $this->assertCount(1, $responseData['messages']['price']);
+
+        $exportedItems = $connection->fetchAll("SELECT export_status FROM s_plugin_connect_items WHERE source_id = '2' AND export_status = 'update' ");
+        $this->assertCount(1, $exportedItems);
```
keulinho: this would still be green if product with source_id 2-123 is marked 'update' despite the error
maybe it would be suficient to remove the WHERE source_id = 2

otherwise you could check the status for 2-123 and assert that it is not update or better that it is error

Tests/Functional/Controller/ImportTest.php
```diff
@@ -71,26 +56,41 @@ public function test_assign_remote_to_local_category_action_without_params()
...
+        $client = self::createBackendClient();
+
+        $connection = self::getDbalConnection();
 
         $this->importFixturesFileOnce(__DIR__ . '/../_fixtures/categories.sql');
 
+        $localCategoryId = '140809703';
+        $remoteCategoryKey = '/deutsch/television';
+
+        $remoteCategoryId = $connection->fetchColumn('SELECT id FROM s_plugin_connect_categories WHERE category_key = ?', [$remoteCategoryKey]);
```
keulinho: IDK whats in the fixture but i think that there's the shopID 1234 for that remote category
so please use it in this query too because the category key alone isn't unique anymore

Tests/Functional/Controller/ImportTest.php
```diff
@@ -217,22 +217,30 @@ public function test_unassign_remote_to_local_category_without_category_id()
     public function test_unassign_remote_articles_from_local_category()
     {
         /** @var TestClient $client */
-        $client = $this->createBackendClient();
+        $client = self::createBackendClient();
```
keulinho: please insert fixtures first and assert before calling the controller that there are assignements
I'm pretty sure the route requires some parameters and what you are actual testing is that it doesn't fail if called without 
needed parameters but the function itself does nothing i guess

Tests/Functional/Controller/ImportTest.php
```diff
@@ -217,22 +217,30 @@ public function test_unassign_remote_to_local_category_without_category_id()
...
+        $this->assertEmpty($categoryAssignments);
     }
 
     public function test_recreate_remote_categories()
     {
         /** @var TestClient $client */
-        $client = $this->createBackendClient();
+        $client = self::createBackendClient();
 
         $client->request('POST', 'backend/Import/recreateRemoteCategories');
```
keulinho: same as above
this does not test what the function is actually doing
please add some fixtures and assertion based on them
here you need as fixture an entry in connect items with some categories but no connect categories
after the call the connect categories should be there

Tests/InitResourceDbSubscriber.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/**
+ * (c) shopware AG <info@shopware.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ShopwarePlugins\Connect\Tests;
+
+class InitResourceDbSubscriber implements \Enlight\Event\SubscriberInterface
```
keulinho: what kind of magic is this?? :D
please add a small comment why this is necessary and what will be broken if it is removed

Tests/KernelTestCaseTrait.php
```diff
@@ -46,9 +45,7 @@
     /**
      * @var array
      */
-//    private $commonTestFixtures = [
-//        __DIR__ . '/../../common_user_test_fixtures.sql',
-//    ];
+    private $commonTestFixtures = [];
```
keulinho: is it really needed or can we simply delete it

PR #415, opened 2017-11-17
CON-4756: CrossSelling infos get synced

keulinho:


composer.json
```diff
@@ -5,7 +5,7 @@
   },
   "require": {
     "firebase/php-jwt": "^3.0",
-    "shopware/connect-sdk": "dev-master#v2.0.6"
+    "shopware/connect-sdk": "dev-CON-4756"
```
keulinho: just for testing and QA has to be changed to official release if SDK 2.0.7 is released

PR #424, opened 2017-11-24
CON-4965: product to local category assignment is aware of streams

keulinho:
Problem was that we don't know wich category from which stream was
assigned to a local category
this made problems in assigning and in ToShop as well
now we can distinguish between different streams


keulinho:
drop of coverage is mostly because new model class that contains just getter and setter and is completely untested

Tests/Integration/Components/CategoryResolver/AutoCategoryResolverTest.php
```diff
@@ -61,7 +61,7 @@ public function testResolve()
         $this->createRemoteCategories($createdCategories);
 
         if (!$bootsCategory) {
-            $this->categoryResolver->createLocalCategory('Boots', '/deutsch/boots', $defaultCategory->getId(), 1234);
+            $this->categoryResolver->createLocalCategory('Boots', '/deutsch/boots', $defaultCategory->getId(), 1234, 'test');
```
sbofirov: I think more speaking name for stream is better, because right now we know it, but after some months string 'test' means every think. 'Boots stream' for example is better.

Tests/Integration/Components/CategoryResolver/AutoCategoryResolverTest.php
```diff
@@ -79,7 +79,7 @@ public function testResolve()
         ]);
 
         if (!$nikeBootsCategory) {
-            $this->categoryResolver->createLocalCategory('Boots', '/deutsch/nike/boots', $nikeCategoryId, 1234);
+            $this->categoryResolver->createLocalCategory('Boots', '/deutsch/nike/boots', $nikeCategoryId, 1234, 'test');
```
sbofirov: same here and all other places where we have 'test' for stream name. My opinion is that we need to use similar data with production, as much as possible.

Tests/Integration/Components/CategoryResolver/DefaultCategoryResolverTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->manager->getConnection()->executeQuery('DELETE FROM s_plugin_connect_categories');
+        $this->manager->getConnection()->executeQuery('DELETE FROM s_plugin_connect_categories_to_local_categories');
+        $this->importFixtures(__DIR__ . '/../_fixtures/categories.sql');
+        $categories = [
+            '/deutsch' => 'Deutsch',
+            '/deutsch/test1' => 'Test 1',
+            '/deutsch/test2' => 'Test 2'
+        ];
+
+        $localCategories = $this->categoryResolver->resolve($categories, 1234, 'Teststream');
```
sbofirov: Teststream is fine. I mean better than just test

Tests/Integration/Components/ImportServiceTest.php
```diff
@@ -277,67 +276,48 @@ public function testFindRemoteArticleIdsByCategoryId()
...
-            'INSERT IGNORE INTO `s_plugin_connect_product_to_categories` (`articleID`, `connect_category_id`) VALUES (?, ?)',
-            [14471, $fantasyCategoryId]
-        );
-
-        // insert invalid articleId in s_plugin_connect_product_to_categories
-        $this->manager->getConnection()->executeQuery(
-            'INSERT IGNORE INTO `s_plugin_connect_product_to_categories` (`articleID`, `connect_category_id`) VALUES (?, ?)',
-            [9087041234, $fantasyCategoryId]
-        );
+        $this->importFixtures(__DIR__ . '/_fixtures/one_item.sql');
```
sbofirov: one_item sounds too generic. I think will be better to use article name or category name if the file contains only one item.

Tests/Integration/Components/_fixtures/simple_connect_items.sql
```diff
@@ -1,28 +1,32 @@
...
-VALUES (14470, 7091851, '14470-7091851', "hash", 123, 1);
-INSERT INTO s_articles_details (id, `articleID`, `ordernumber`, `kind`) VALUES (7091852, 14470, 'sw1007.3', 2);
-INSERT INTO s_plugin_connect_items (article_id, article_detail_id, source_id, purchase_price_hash, offer_valid_until, stream)
-VALUES (14470, 7091852, '14470-7091852', "hash", 123, 1);
\ No newline at end of file
+INSERT INTO s_articles (`id`, `name`, main_detail_id, changetime, pricegroupActive, laststock, crossbundlelook, notification, template, mode)
+VALUES (14470, 'Tom Joule Wellibob', 7091850, NOW(), 1, 1, 0, 0, 1, 0);
+INSERT INTO s_articles_details (id, `articleID`, `ordernumber`, `kind`, `position`) VALUES (7091850, 14470, 'sw1007', 1, 0);
+INSERT INTO s_plugin_connect_items (article_id, article_detail_id, source_id, purchase_price_hash, offer_valid_until, stream, shop_id)
+VALUES (14470, 7091850, 14470, "hash", 123, "Awesome products", NULL);
```
sbofirov: :+1:

Tests/Unit/Models/RemoteCategoryTest.php
```diff
@@ -0,0 +1,63 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace ShopwarePlugins\Connect\Tests\Unit\Components;
+
+use Shopware\CustomModels\Connect\RemoteCategory;
+use Shopware\CustomModels\Connect\RemoteToLocalCategory;
+use Shopware\Models\Category\Category;
+
+class RemoteCategoryTest extends \PHPUnit_Framework_TestCase
```
sbofirov: Do we need dedicated test for entity ? I think we can just make assertions where we test ```CategoryResolver::storeRemoteCategories```

Tests/Unit/Models/RemoteToLocalCategoryTest.php
```diff
@@ -0,0 +1,60 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace ShopwarePlugins\Connect\Tests\Unit\Components;
+
+use Shopware\CustomModels\Connect\RemoteCategory;
+use Shopware\CustomModels\Connect\RemoteToLocalCategory;
+use Shopware\Models\Category\Category;
+
+class RemoteToLocalCategoryTest extends \PHPUnit_Framework_TestCase
```
sbofirov: same here. I think will be better if we can make assertions in some integration test.

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -14,21 +14,21 @@ class DefaultCategoryResolver extends CategoryResolver
...
-            }
-        }
-
-        return array_map(function ($category) {
-            return $category->getId();
-        }, $localCategories);
+        return $this->manager->getConnection()->executeQuery('
+            SELECT local_category_id
+            FROM s_plugin_connect_categories_to_local_categories
+            WHERE (stream = ? OR stream IS NULL) AND remote_category_id IN (?)',
```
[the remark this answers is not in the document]
keulinho: good point
stream = NULL is for all the existing assignments because of backwards compatibility

Tests/Unit/Models/RemoteCategoryTest.php
keulinho: we make that tests 
i added them because test coverage did go down

Tests/Unit/Models/RemoteToLocalCategoryTest.php
keulinho: we have integration test for most of it

keulinho: problem is currently that we try to use less and less doctrine entities because of performance so some of this isn't used anymore

PR #428, opened 2017-11-28
CON-4976 Always apply variant properties

sbofirov:
If product doesn't have variants and after that at some point they're received it should be configured in merchant shop.


Components/ProductToShop.php
```diff
@@ -508,8 +506,7 @@ private function checkIfMainVariant(DetailModel $detail, DetailModel $mainDetail
      */
     private function updateConfiguratorSetTypeFromProduct(ProductModel $model, Product $product)
     {
-        if (!empty($product->variant)) {
-            $configSet = $model->getConfiguratorSet();
+        if (!empty($product->variant) && $configSet = $model->getConfiguratorSet()) {
```
Christian-Rades: Maybe a seperate check if getConfiguratorSet is null is better style

PR #433, opened 2017-12-04
CON-4973: subcategories in default category resolver are created

keulinho:
before we don't create new subcategories if the parent connect category
is assigned
now it is checked that all subcategories of each assigned remote
category are there and if not they are created

### DIFF
--- a/Components/CategoryResolver/DefaultCategoryResolver.php
+++ b/Components/CategoryResolver/DefaultCategoryResolver.php
@@ -48,13 +48,19 @@ class DefaultCategoryResolver extends CategoryResolver
      */
     private function getDirectlyAssignedCategories($remoteCategoryId, $shopId, $stream)
     {
-        return (array) $this->manager->getConnection()->executeQuery('
+        $localCategories = $this->manager->getConnection()->executeQuery('
               SELECT pclc.local_category_id
               FROM s_plugin_connect_categories_to_local_categories AS pclc
               INNER JOIN s_plugin_connect_categories AS pcc ON pcc.id = pclc.remote_category_id
               WHERE pcc.category_key = ? AND pcc.shop_id = ? AND (pclc.stream = ? OR pclc.stream IS NULL) ',
             [$remoteCategoryId, $shopId, $stream]
         )->fetchAll(\PDO::FETCH_COLUMN);
+
+        if ($localCategories === false) {
+            return [];
+        } else {
+            return $localCategories;
+        }
     }
 
     /**
--- a/Components/CategoryResolver/DefaultCategoryResolver.php
+++ b/Components/CategoryResolver/DefaultCategoryResolver.php
@@ -29,33 +29,69 @@ class DefaultCategoryResolver extends CategoryResolver
      */
     private function findAssignedLocalCategories(array $children, $shopId, $stream)
     {
-        $childCategories = [];
         $mappedCategories = [];
         foreach ($children as $category) {
-            $localCategories = (array) $this->manager->getConnection()->executeQuery('
+            $localCategories = $this->getDirectlyAssignedCategories($category['categoryId'], $shopId, $stream);
+            //use + not array_merge because we want to preserve the numeric keys
+            $mappedCategories = $mappedCategories + $this->buildLocalSubtrees($shopId, $stream, $localCategories, $category['children']);
+            $mappedCategories = $mappedCategories + $this->findAssignedLocalCategoriesForChildCategories($shopId, $stream, $category['children']);
+        }
+
+        return $mappedCategories;
+    }
+
+    /**
+     * @param int $remoteCategoryId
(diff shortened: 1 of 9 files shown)


Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -16,26 +16,48 @@ class DefaultCategoryResolver extends CategoryResolver
...
+        foreach ($children as $category) {
+            $localCategories = $this->manager->getConnection()->executeQuery('
+              SELECT pclc.local_category_id
+              FROM s_plugin_connect_categories_to_local_categories AS pclc
+              INNER JOIN s_plugin_connect_categories AS pcc ON pcc.id = pclc.remote_category_id
+              WHERE pcc.category_key = ? AND pcc.shop_id = ? AND (pclc.stream = ? OR pclc.stream IS NULL) ',
+              [$category['categoryId'], $shopId, $stream]
+            )->fetchAll(\PDO::FETCH_COLUMN);
+
+            if ($localCategories) {
```
JanPietrzyk: replace this with an array cast

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -16,26 +16,48 @@ class DefaultCategoryResolver extends CategoryResolver
...
 
     /**
-     * {@inheritdoc}
+     * @param array $children
+     * @param int $shopId
+     * @param string $stream
+     * @return array
      */
-    public function generateTree(array $categories, $idPrefix = '')
+    private function traverseTree(array $children, $shopId, $stream)
```
JanPietrzyk: how come the name change? traverse with what?

keulinho: generate Tree was an abstract function before 
the defaultCategoryResolver did just return an empty array in that function
that's why i extracted the logic from autoCategoryResolver to the base class

keulinho:
drop of coverage is caused by using phpdbg instead of xdebug
this was merged into master earlier today

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -16,26 +16,48 @@ class DefaultCategoryResolver extends CategoryResolver
...
+        foreach ($children as $category) {
+            $localCategories = $this->manager->getConnection()->executeQuery('
+              SELECT pclc.local_category_id
+              FROM s_plugin_connect_categories_to_local_categories AS pclc
+              INNER JOIN s_plugin_connect_categories AS pcc ON pcc.id = pclc.remote_category_id
+              WHERE pcc.category_key = ? AND pcc.shop_id = ? AND (pclc.stream = ? OR pclc.stream IS NULL) ',
+              [$category['categoryId'], $shopId, $stream]
+            )->fetchAll(\PDO::FETCH_COLUMN);
+
+            if (is_array($localCategories)) {
```
JanPietrzyk: I meant the entire if is redundant:

```php
$data = (array) $this->giveMeAnArrayOrNull();

foreach($data as $key => $value) {
   ...
}
```
in this case the $data will always be an array

* `giveMeAnArrayOrNull` returns an array => $data is untouched
* `giveMeAnArrayOrNull` returns null => $data is an empty array with a count of 0

JanPietrzyk:
@keulinho Can you please do me a favor and mark the class `Components/CategoryResolver`with a todo to refactor it? Other then that: Good work!

Components/CategoryResolver/DefaultCategoryResolver.php
keulinho: does this work even is $data is false?
i dont know if there are differences between the different hydration modes 
but Connection::fetchColumn() will return false if the query doesn't match anything
that was the intent behind the original if

Components/CategoryResolver/DefaultCategoryResolver.php
```diff
@@ -16,26 +16,46 @@ class DefaultCategoryResolver extends CategoryResolver
...
+        foreach ($children as $category) {
+            $localCategories = (array) $this->manager->getConnection()->executeQuery('
+              SELECT pclc.local_category_id
+              FROM s_plugin_connect_categories_to_local_categories AS pclc
+              INNER JOIN s_plugin_connect_categories AS pcc ON pcc.id = pclc.remote_category_id
+              WHERE pcc.category_key = ? AND pcc.shop_id = ? AND (pclc.stream = ? OR pclc.stream IS NULL) ',
+              [$category['categoryId'], $shopId, $stream]
+            )->fetchAll(\PDO::FETCH_COLUMN);
+
+            foreach ($localCategories as $localCategory) {
```
JanPietrzyk: you were correct https://3v4l.org/poOLF, so please move the query to a seperate method that guarantees that it returns an array

PR #443, opened 2017-12-19
[CON-4517] move import preferences to import tab

HoelShare:


Views/backend/connect/view/config/import/form.js
```diff
@@ -299,7 +310,61 @@ Ext.define('Shopware.apps.Connect.view.config.import.Form', {
             ]
         });
 
-        return [ containerTop, containerMiddle, containerBottom ];
+        elements.push(productContainer);
+        elements.push(imageContainer);
+        elements.push(orderContainer);
+        return elements;
```
tobiasberge: Not sure right now what Ext.create() returns but maybe this can be simplified with:
```

elements.push(productContainer, imageContainer, orderContainer);
```

PR #446, opened 2018-01-02
CON-4850: used SDK functionality to check if priceType is valid

keulinho:


composer.json
```diff
@@ -5,7 +5,7 @@
   },
   "require": {
     "firebase/php-jwt": "^3.0",
-    "shopware/connect-sdk": "dev-master#v2.0.8"
+    "shopware/connect-sdk": "dev-CON-4850"
```
HoelShare: REMINDER: Please change back, when the PR of the sdk is merged.

PR #447, opened 2018-01-02
[CON-4522] Move drop shipping options to import tab

HoelShare:
This PR depends on PR #443


Bootstrap/Update.php
```diff
@@ -627,4 +628,20 @@ private function addOverwriteMainImage()
             }
         }
     }
+
+    private function changeGroupNameImportSettings()
+    {
+        if (version_compare($this->version, '1.1.12', '<=')) {
```
keulinho: `$this->version`contains the old version so when updating from 1.1.11 to 1.1.12 it would contain 1.1.11 
i think you have to check less or eguals '1.1.11'
in general it is Version_with_that_change - 1

plugin.json
```diff
@@ -10,6 +10,7 @@
     "currentVersion": "1.1.11",
     "changelog": {
         "de": {
+            "1.1.12": "CON-4522: Drop shipping Einstellungen in den Import Tab verschoben",
```
keulinho: in general i would not add the changelog in this moment because you never really know with wich release your PR is realesed.
The workflow is to flag the jira ticket with the plugin version you think and then when we want to do  package test we add al changelogs for that release

PR #448, opened 2018-01-03
CON-4632

sbofirov:


sbofirov:
There is another PR in Connect-SDK repository -> https://github.com/shopware/Connect-SDK/pull/41

Components/ProductFromShop.php
```diff
@@ -407,14 +408,31 @@ public function updatePaymentStatus(PaymentStatus $status)
     public function calculateShippingCosts(Order $order)
     {
         if (!$order->deliveryAddress) {
-            return new Shipping(['isShippable' => false]);
+            return new Shipping([
+                'isShippable' => false,
+                'messages' => [
+                    new Message([
+                        'message' => 'Delivery address could not be empty'
```
sbofirov: all english messages should be translated to german or completely replaced by other messages.

Components/ProductFromShop.php
```diff
@@ -443,7 +468,14 @@ public function calculateShippingCosts(Order $order)
         // todo: if products are not shippable with default shipping
         // todo: do we need to check with other shipping methods
```
keulinho: what about these todos?

composer.json
```diff
@@ -5,7 +5,7 @@
   },
   "require": {
     "firebase/php-jwt": "^3.0",
-    "shopware/connect-sdk": "dev-master#v2.0.8"
```
keulinho: reminder to change back if SDK is trelesed ;)

sbofirov:
REMINDER: Update SDK to latest master after merging SDK PR.

Components/ProductFromShop.php
```diff
@@ -500,6 +481,39 @@ public function calculateShippingCosts(Order $order)
...
+                        'message' => 'order_not_shippable_to_country',
+                        'values' => [
+                            'country' => $countryIso3,
+                        ]
+                    ])
+                ]
+            ]);
+        }
+
+        $shippingMethod = reset($shippingMethods);
```
keulinho: how are these shippingcosts ordered? because we just take the first but doesn't it make sense to take the cheapest?

Views/backend/connect/view/config/import/form.js
```diff
@@ -484,7 +484,7 @@ Ext.define('Shopware.apps.Connect.view.config.import.Form', {
                     fieldLabel: me.snippets.separateShippingLabel,
```
keulinho: don't you have to check this value in the basket?

Components/BasketHelper.php
```diff
@@ -466,8 +500,6 @@ public function recalculate(CheckResult $checkResult)
         $this->basket['sShippingcostsNet'] += $shippingCostsNet;
         $this->basket['sShippingcostsWithTax'] += $shippingCostsWithTax;
 
-        // Update total amount
-        $this->basket['AmountNumeric'] += $shippingCostsNet;
```
keulinho: isn't `$this->basket['AmountNumeric'] += $shippingCostsWithTax;` right and neccessary?

keulinho:
I'm just curious what if i order two products in merchant shop and supplier has remote shipping cost do i have to pay 2 times the shipping?
long story short: do we check shipping per product or per supplier?
if the first is the case please create a follow up ticket and we can discuss in planning wether we want to change this

Components/ProductFromShop.php
sbofirov: Shopware will decide which shipping cost method to return by given country and products in basket. You can see in ```Shopware_Controllers_Frontend_Checkout::getSelectedDispatch```

```
/**
     * Get selected dispatch or select a default dispatch
     *
     * @return bool|array
     */
    public function getSelectedDispatch()
    {
        if (empty($this->session['sCountry'])) {
            return false;
        }

        $dispatches = $this->admin->sGetPremiumDispatches($this->session['sCountry'], null, $this->session['sState']);
        if (empty($dispatches)) {
            unset($this->session['sDispatch']);

            return false;
        }

        foreach ($dispatches as $dispatch) {
            if ($dispatch['id'] == $this->session['sDispatch']) {
                return $dispatch;
            }
        }
        $dispatch = reset($dispatches);
        $this->session['sDispatch'] = (int) $dispatch['id'];

        return $dispatch;
    }
```

Views/backend/connect/view/config/import/form.js
sbofirov: If this option is enabled, Connect plugin will show blue label "Separate dispatch" next to the product and separate line in basket, like vouchers.

Components/BasketHelper.php
sbofirov: Some lines below at the end of this method I saw ```$this->basket['AmountNumeric'] = $this->basket['sAmount'];```. This is the reason why I deleted this line.

sbofirov:
If you have more than 1 product from same supplier you have to pay his shipping costs only once.
Let's say you have 4 products in your basket.

1 Local product which is not part of Connect.
2 Connect products imported from Supplier A
1 Connect product imported from Supplier B.

Shipping cost type is remote.

Then in basket you will see 
Once shipping cost from local shop
Once shipping cost from Supplier A
Once shipping cost from Supplier B.

If you have offer with Supplier B where shipping cost type = all
Then as a customer you'll see Local shipping cost and one shipping cost from Supplier A.

Just for info, you can check SDK::checkProducts. Connect plugin makes call to this method with given order. it contains all products from supplier. Then SDK will split to smaller orders for every unique supplier. After every separate checkResult, SDK will combine them in aggregatedCheckResult.

PR #454, opened 2018-02-02
CON-5041 Use latest supported languages in SDK

sbofirov:


composer.json
```diff
@@ -5,7 +5,7 @@
   },
   "require": {
     "firebase/php-jwt": "^3.0",
-    "shopware/connect-sdk": "dev-master#v2.0.11"
+    "shopware/connect-sdk": "dev-CON-5041"
```
keulinho: change that as soon as we have released the sdk

PR #459, opened 2018-02-09
CON-5110 Add local shipping costs when necessary

sbofirov:
Add local shipping costs price even when basket contains only connect products.
This is needed when at least one supplier has shipping cost type = all or filter.
Then endcustomer has to pay merchant shipping costs and shipping costs provided by
other suppliers remotely.

### DIFF
--- a/Components/BasketHelper.php
+++ b/Components/BasketHelper.php
@@ -22,6 +22,8 @@ use Shopware\Connect\Struct\Product;
  */
 class BasketHelper
 {
+    const REMOTE_SHIPPING = 'remote';
+
     /**
      * The basket array decorated by this class
      * @var array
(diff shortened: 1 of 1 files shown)


Components/BasketHelper.php
```diff
@@ -223,9 +225,11 @@ public function fixBasket()
             $this->onlyConnectProducts = true;
 
             $this->removeNonProductsFromBasket();
-            $this->basket = $this->removeDefaultShipping($this->basket);
 
             $connectContent = $this->getConnectContent();
+            if (false === $this->customerHasToPayLocalShipping($connectContent)) {
```
keulinho: image
Please don't use yoda conditions ;)

svenfinke: In fact, this was a defined and wanted code style back when QaFoo was on the project, and in case of false or null there are good reasons to do it this way.

If you write false at the beginning of the statement, it is easy to scan the code for such conditions or to locate wrong logic. It is also the preferred way instead of writing "if(!$comething)" because the "!" is easy to miss.

So, use this for "false" and "null", but not for other constants.

Components/BasketHelper.php
```diff
@@ -239,6 +243,33 @@ public function fixBasket()
         return false;
     }
 
+    /**
+     * Verifies that some of suppliers as shippingCosts type
```
SebastianFranze: there is a little typo: "as" should be "has" schouldn't it?

PR #465, opened 2018-02-26
[CON-5085] Make connect plugin compatible with smarty security

SebastianFranze:
template registration is now called at `Enlight_Controller_Action_PreDispatch`event

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -7,9 +7,10 @@
     "license": "proprietary",
     "link": "http://store.shopware.com",
     "author": "Shopware AG",
-    "currentVersion": "1.1.14",
+    "currentVersion": "1.1.15",
     "changelog": {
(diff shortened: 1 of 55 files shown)


plugin.json
```diff
@@ -7,9 +7,10 @@
     "license": "proprietary",
     "link": "http://store.shopware.com",
     "author": "Shopware AG",
-    "currentVersion": "1.1.14",
+    "currentVersion": "1.1.15",
     "changelog": {
         "de": {
+            "1.1.15": "CON-5085 - Templates werden jetzt früher im Dispatchprozess registriert",
```
keulinho: yep that s the technical description but users don't understand this 
i would recommend to change this to something like
`Added: Compatibility for SmartySecurity`

PR #473, opened 2018-04-12
[CON-5195] user is now notified why a order can't be delivered

SebastianFranze:


composer.json
```diff
@@ -5,7 +5,7 @@
   },
   "require": {
     "firebase/php-jwt": "^3.0",
-    "shopware/connect-sdk": "dev-master#v2.0.12"
+    "shopware/connect-sdk": "dev-CON-5195"
```
SebastianFranze: Don't merge until sdk is released

Snippets/frontend.sql
```diff
@@ -66,5 +66,6 @@ ON DUPLICATE KEY UPDATE
   `namespace` = VALUES(`namespace`),
   `shopID` = VALUES(`shopID`),
   `name` = VALUES(`name`),
-  `localeID` = VALUES(`localeID`)
+  `localeID` = VALUES(`localeID`),
+  `value` = VALUES(`value`)
```
SebastianFranze: I don't know if this is a problem. Users could overwrite our snippets. With this new Update we would overwrite it back to our snippets every time the plugin is installed (updated)

Components/BasketHelper.php
```diff
@@ -664,6 +668,24 @@ public function getConnectTemplateVariables(array $connectMessages)
         ];
     }
 
+    private function guessProductTitle(Message $message)
```
JanPietrzyk: mach ma `determineProductTitles`, wa?

Snippets/frontend.sql
JanPietrzyk: weil der Rest der Datei ja ein ähnliches Problem hat, find ich das zwar Schade, aber nicht soooo schlimm. Hat es einen Grund warum Connect die Snippets nicht in den ini-Files hat?

